PR #24, opened 2015-01-19
added medium tau id

mstoev:


peiffer:
The deltaR cut does not work! Please fix it by adding a "return false" if deltaRmin<0.5.

PR #25, opened 2015-01-19
added medium tau id

mstoev:


peiffer:
There is still a bug. Please fix it!

PR #64, opened 2015-03-27
adjust drmax to ak4 radius

HeinAtCERN:
Dear Thomas, 
please check that this makes sense. Physically. =)

Best, -H


peiffer:
The change makes sense for ak4 jets. Please also update the documentation https://github.com/HeinAtCERN/UHH2/blob/2c150c63b0f9d10b8a42e7c4ef9f41e6ad36eb80/common/include/JetCorrections.h#L103

PR #70, opened 2015-04-15
added singletops in next-ntuple-format

reimersa:


rkogler:
Please use absolute path names, such that other users can use these files as well. In general, the path should look like /<dir1>/<dir2>/<dir3>/.... without two dots. Usually, we have the files in /pnfs/desy.de/cms/tier2/store/user/<username>/<dir>. Could you please put them there?

PR #73, opened 2015-04-23
util function and ST histogram

HeinAtCERN:


peiffer:
this is a spacelike-save MASS function, not deltaR

HeinAtCERN:
Right! =)

PR #75, opened 2015-04-28
reproduced LQtoTopMu ntuples

mstoev:
Not all the muons have been collected by the filter on genlvl, so I reproduced the ntuples


peiffer:
Can you also make a pull request for the changed genparticle pruner in the ntuplewriter.py script?

PR #77, opened 2015-05-12
addition of a ttbar-chi2 discriminator for top-tagged events

missirol:
* addition of an alternative chi2-discriminator (in ReconstructionHypothesisDiscriminators), to be used for events with a top-tagged jet
  * hadronic term of the chi2 calculated using the groomed (instead of ungroomed) topjet mass

* added a pointer to TopJet as data member of the ReconstructionHypothesis class, in order to be able to access TopJet-specific methods (e.g. subjets)


peiffer:
Hi Marino,
can you announce once again on the mailing list that this change has now been included?

PR #79, opened 2015-05-14
undeclare event output

HeinAtCERN:
declaring and implementing Context::undeclare_event_output(string), to remove branches from output.
Implementation in CMSSWContext simply throws a runtime exception.


peiffer:
Can you also commit an example how to apply undeclare_event_output, perhaps you can modify ExampleModuleEventOutput.cxx?

PR #81, opened 2015-05-17
Added batch framework submission for FNAL

drberry85:
I added the batch submission python framework we use here at FNAL on condor. It has been updated to UHH2.

Doug


peiffer:
Can you also submit this commit into the next-ntuple-format branch?

PR #95, opened 2015-06-11
updates for lepton reconstruction (mini-isolation)

missirol:
* addition of some variables for lepton reconstruction
 * mini-isolation
 * muon medium-ID
 * electron MVA

* important: updates tested in CMSSW_7_4_3 (on top of the compilation instructions in the UHH2/ twiki)

* mini-isolation:
 * mini-isolation deposits are calculated using 2 plugins (located in core/plugins/):

  * MUONS: plugin="CandIsolatorFromDepositsMINIISO": minimal generalization of the standard plugin "PhysicsTools/IsolationAlgos/plugins/CandIsolatorFromDeposits.*", which is used for the standard (fixed-cone) calculation of the isolation deposits for muons (see RecoMuon.MuonIsolation.muonPFIsolationValues_cff).

  * ELECTRONS: plugin="PFCandIsolatorFromDepositsMINIISO": minimal generalization of the standard plugin "CommonTools/ParticleFlow/plugins/PFCandIsolatorFromDeposit.*", which is used for the standard (fixed-cone) calculation of the isolation deposits for muons (see RecoParticleFlow.PFProducer.electronPFIsolationValues_cff).

 * value-maps for the mini-isolation deposits are loaded in core/python/*_pfMiniIsolation_cff.py
  * this is done in a way that (aside for handling the cone size) the configuration of the standard modules for the isolation deposits is used (minimum pt threshold on the candidates, geometrical vetos etc)
  * the 3 parameters for the mini-isolation cone can be configured from arguments of the python function. the default values are those proposed in the SUSY PAG from their first studies on mini-isolation (see https://indico.cern.ch/event/388718/contribution/5/material/slides/0.pdf).

 * 6 values are computed for each lepton: sum_pT deposits for charged-hadrons ("CH"), neutral-hadrons ("NH"), photons ("Ph"), charged pileup ("PU"), neutral-hadrons with PF-reweighting ("NH_pfwgt"), photons with PF-reweighting ("Ph_pfwgt"). PF-reweighting is a new method proposed for PU mitigation (see, for example, https://github.com/cms-sw/cmssw/pull/4131 and refs therein)
 * added new float members to Muon and Electron classes for mini-iso sum-pT deposits

* muon medium-ID:
 * value taken from pat::Muon::isMediumMuon()
 * added via the tag mechanism in Muon class

* electron MVA:
 * added via to collection of pat::Electron via the addUserFloat method
 * implementation taken from example in https://github.com/HuguesBrun/ExampleElectronMVAid/blob/addMVAfromMiniAOD/plugins/ExampleElectronMVAid.cc (as https://twiki.cern.ch/twiki/bin/view/CMS/MultivariateElectronIdentificationRun2)
 * added to Electron class through the (already present) float members "mvaNonTrigV0" and "mvaTrigV0"
 * .xml files for the MVA calculation have to be downloaded by hand (as far as i know) before running. a script that does that can be found under core/env.sh

### DIFF
--- a/core/include/Electron.h
+++ b/core/include/Electron.h
@@ -4,19 +4,25 @@
 #include "Tags.h"
 #include "FlavorParticle.h"
 
-class Electron: public Particle{
-public:
-    
-  enum tag {eid_PHYS14_20x25_veto, eid_PHYS14_20x25_loose, eid_PHYS14_20x25_medium, eid_PHYS14_20x25_tight };
-  
+class Electron: public Particle {
+
+ public:
+  enum tag {
+    eid_PHYS14_20x25_veto,
+    eid_PHYS14_20x25_loose,
+    eid_PHYS14_20x25_medium,
+    eid_PHYS14_20x25_tight,
+  };
+
   static tag tagname2tag(const std::string & tagname){
-      if(tagname == "eid_PHYS14_20x25_veto") return eid_PHYS14_20x25_veto;
-      if(tagname == "eid_PHYS14_20x25_loose") return eid_PHYS14_20x25_loose;
-      if(tagname == "eid_PHYS14_20x25_medium") return eid_PHYS14_20x25_medium;
-      if(tagname == "eid_PHYS14_20x25_tight") return eid_PHYS14_20x25_tight;
-      throw std::runtime_error("unknown Electron::tag '" + tagname + "'");
+    if(tagname == "eid_PHYS14_20x25_veto")   return eid_PHYS14_20x25_veto;
+    if(tagname == "eid_PHYS14_20x25_loose")  return eid_PHYS14_20x25_loose;
+    if(tagname == "eid_PHYS14_20x25_medium") return eid_PHYS14_20x25_medium;
+    if(tagname == "eid_PHYS14_20x25_tight")  return eid_PHYS14_20x25_tight;
+
+    throw std::runtime_error("unknown Electron::tag '" + tagname + "'");
   }
-  
+
   Electron(){
    m_supercluster_eta = 0; 
    m_supercluster_phi = 0; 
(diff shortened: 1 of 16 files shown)


peiffer:
Hi Marino,

thanks for this update. But do we have to include all these plugins for the mini isolation into our own code? Are they not available somewhere in CMSSW, even in a private fork from where we could download them (which would have to be added to the installation instructions)?

missirol:
Hi Thomas,

the two plugins for the mini-isolation deposits are merely copies of the corresponding plugins in CMSSW that are used for the fixed-cone isolation (see the PR header for details).
the only difference is the one line where the isolation cone is calculated based on the lepton pT (i did not find a way to make this generalization work by class inheritance).

as far as i know, there are no official or partial recipes for this as of now, that's what i thought we were discussing at the end of the last meeting.

the only thing i could find is a reference given in a SUSY hypernews (https://hypernews.cern.ch/HyperNews/CMS/get/susy/1991.html), where the mini-iso vars are calculated inside an ntuplemaker and several conditions like the pt-threshold and the cone-vetos are hard-coded.
i decided to use a more modular approach, also relying as much as possible on the standard CMSSW modules for internal parameters like pt-thresholds and cone-vetos (which, like this, should be automatically aligned to the central cmssw release). that implies the addition of two plugins for mini-isolation of muons and electrons, as long as official ones are missing.

of course, if an official recipe comes out and is integrated in CMSSW, we can simply remove these additions and align ourselves to the recommendations.
i understood we wanted to avoid waiting and be able to test something already.
if we prefer waiting for the official recipe on mini-isolation, it's fine.

Marino

peiffer:
Hi Marino,

okay, fine for me. I just thought, that the susy group might have something more flexible already. But if that is not the case we can keep the plugins in our code until there is an official CMSSW module.

Thomas

PR #100, opened 2015-06-29
make PFHT800 emulation independent of PFHTForMC version

emanueleusai:
Over concerns that the PFHTForMC trigger path used to emulate PFHT800 might change version across different MC productions I made the code independent the trigger version.


imarches:
thanks

PR #117, opened 2015-07-21
JetHT dataset ntuplized

emanueleusai:
Added JetHT dataset, DCS-only lumi filter used


peiffer:
Could you please add this samples to the wiki page https://github.com/UHH2/UHH2/wiki/Ntuple-Prodcution-%28Run-II,-1st-round,-50ns%29 ?

emanueleusai:
Done! I forgot about that, thanks for reminding me

PR #126, opened 2015-07-30
inclusive QCD Pt binned from pthat=15GeV to infinity

emanueleusai:
In principle two low energy samples are missing to complete the QCD pT binned sample:
10<ptHat<15 GeV
5<ptHat<10 GeV
but that are so low in energy that I'm not planning to process them unless there is a specific need.
Note that a minor number of events in the lower pT samples have a problem with PFHT trigger emulation. Write me for a possible solution to the issue. Eventually I might think of rerunning those with a fixed version of the ntuplewriter.


imarches:
We are most probably going to have a new round of production relatively soon, so I would probably wait for this before re-running anything

PR #128, opened 2015-07-30
Bug in event.genInfo->weights

dgonzal:
I observe a bug in the way this is filled for at least the WJets sample and also other samples seem to be affected. 
event.genInfo->weights can be negativ and for W+Jets it is   -/+225892 (negativ ?!)
If I normalise my PreSelection and my Selection the weight is multiplied such that if the weight in the PreSelection [0,1] the values become smaller & if 1> the weight are higher.
This need some more investigation. This is just an shortterm solution


peiffer:
Hi Daniel,

this is not a bug, this is a feature of the aMC@NLO sample. You have to take these weights into account when normalizing distributions. For the absolute normalization, you have to take the sum of all weights instead of the number of events.

Thomas

PR #133, opened 2015-07-31
save full jet information in common/ReconstructionHypothesis

missirol:
* keep the Jet object in the ttbar-hypothesis container
  * useful to access jet variables, e.g. b-tag, in testing methods for the ttbar kinematical reconstruction

### DIFF
--- a/common/include/ReconstructionHypothesis.h
+++ b/common/include/ReconstructionHypothesis.h
@@ -23,8 +23,8 @@ public:
   LorentzVector tophad_v4() const{return m_tophad_v4;} 
   LorentzVector neutrino_v4() const{return m_neutrino_v4;} 
   Particle lepton() const{return m_lepton;}
-  std::vector<Particle> toplep_jets() const{return m_toplep_jets;}
-  std::vector<Particle> tophad_jets() const{return m_tophad_jets;}
+  std::vector<Jet> toplep_jets() const{return m_toplep_jets;}
+  std::vector<Jet> tophad_jets() const{return m_tophad_jets;}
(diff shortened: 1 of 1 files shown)


peiffer:
Isn't the file size increasing a lot when storing the full jet info for all hypotheses? The number of hypotheses might be quite large. If you really need to access the jet info, you can still make a matching of the four-momenta.

missirol:
yes, for sure the size gets bigger.
in general, we should probably move past the idea of saving all the hypotheses in sub-ntuples, because that is going to be demanding in terms of disk-space anyway, even using Particle, especially if one wants to run the analysis more than once testing different things, which is a very likely scenario moving forward.
i think the idea should be having a ttbar-hypothesis with all the information and then pick what to save at the analysis level, instead of just saving everything at the cost of reducing information in each hypothesis.

making a matching a posteriori with the jet collection could probably work, but it s just a workaround, not the right method.

what do you think?

peiffer:
Okay, if we do not store all possible reconstruction hypotheses for every event but just the best one that we will use later, we can certainly store the full jet information in each hypothesis.

missirol:
ok, if this is ok then, i'll prepare a PR to change the default workflow in ZprimeSemileptonic/ZprimeSelectionModule (which currently saves all hypotheses).
after that we can include this change in UHH2/

peiffer:
Fine, I will merge this pull request when you have implemented the changes in ZprimeSemileptonic.

missirol:
UHH2/ZprimeSemiLeptonic#13

PR #163, opened 2015-08-19
added run2-50ns ZprimeToTT samples in common/datasets/RunII_50ns_v2/

missirol:
xml files for ZprimeToTT signal samples for run2-spring15 production with 50ns BX [v2]

* processed with commit 150f11d of UHH2/


missirol:
#164

PR #168, opened 2015-08-20
changed some histograms inTopJetHists, added TopJetIds: HepTopTagV2, Type2TopJet, Tau2/Tau1

marmeyer:
For the HepTopTagV2 TopJetId I used the cuts for the 40% efficiency WP given in the talk about top tagging in this B2G meeting: https://indico.cern.ch/event/393337/#preview:1078843. The cuts have to be updated when there are new recommendations.


peiffer:
Hi Mareike, 
could you write some documentation what the HepTopTag and the Type2TopTag are doing into the header file?

PR #169, opened 2015-08-20
QCD dataset from Pt600 to PtInf

HeinAtCERN:


peiffer:
Could you put new xml files into the directory common/datasets/RunII_50ns_v2 instead of replacing the old files? We can delete the v1 xml files when all v2 samples are in place.

PR #196, opened 2015-09-08
changed TopJetCorrector: now one can choose which topjet collection should be corrected (through a handle)

marmeyer:


emanueleusai:
Wasn't this functionality already implemented in GenericTopJetCorrector?

peiffer:
Oops, you are right, we somehow overlooked that.

PR #203, opened 2015-09-21
updates for common/ElectronIds [UHH2:RunII_50ns]

missirol:
* updated thresholds for electron cut-based ID
  * https://hypernews.cern.ch/HyperNews/CMS/get/egamma/1623.html
  * https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedElectronIdentificationRun2

* restructured the implementation of the ElectronID selectors in common/\*/ElectronIds.\*

  * these updates are backward-compatible (only the underlying implementation has changed)

  * the idea is to simplify the hard-coding of thresholds and make the implementation (as much as possible) not error-prone

  * the new implementation complies with the "table" structure that the Egamma twikis use for the ID working points

   * this allows to update/add ID recipes more easily, just copying the twiki table all at once in one place (instead of hard-coding thresholds 1-by-1 in different places)

   * also, the implementation is done so that the cut-based-ID thresholds are hard-coded only once (and not twice, as before, for cut-based IDs with and without isolation cut)

### DIFF
--- a/common/include/ElectronIds.h
+++ b/common/include/ElectronIds.h
@@ -1,6 +1,9 @@
 #pragma once
 
-#include "UHH2/core/include/Event.h"
+#include <unordered_map>
+#include <vector>
+
+#include <UHH2/core/include/Event.h>
 
 // Note that the electronIds are implemented not by inheriting from a common base class, but
 // rather as 'something' that can be used as function with the signature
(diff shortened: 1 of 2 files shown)


rkogler:
Nice work! Did you test the performance and implementation?

missirol:
yes, i have ran with these changes for a few days now and they work in my tests.
the performance seems comparable to the previous implementation (or at least, it affects as little the overall performance). using the ZprimePreSelectionModule in proof/lite with 10 nodes on a signal sample (zprime-3TeV), i got the same execution time (11s).

of course, if somebody reports any problem related to this update we can switch back (the physics is the same, this implementation is just clearer, at least to me)

rkogler:
Yes, the implementation is very nice and clear now. It's also easy for updates with new recommendations. 
Thanks for testing it, I was wondering about the performance of the value maps with strings. It's good that you don't see a performance deterioration. I merged the request.

PR #221, opened 2015-10-13
Added RunII_25ns_v1 Background Monte Carlo

drberry85:
Thomas,

Hopefully the branches match this time.

Doug


peiffer:
Thanks, that looks better.

Thomas

imarches:
Thanks Doug for taking care of all these MC datasets. It would be nice if you could update the information here:

https://github.com/UHH2/UHH2/wiki/Ntuple-Production-(Run-II,-25ns-v1-MC-ONLY!)

when you have time.

PR #222, opened 2015-10-13
more info in context

HeinAtCERN:
Will this eventually be merged into master again?


HeinAtCERN:
Maybe a bit of explanation:
The reason for adding this is to check if all events have been processed. If not all events were processed (if some input files are corrupted or so), the number is needed to calculate a factor for fixing the histogram normalizations.

peiffer:
I have added this commit to the master branch, see #223

PR #248, opened 2015-11-06
remove MET corrections

peiffer:
I noticed that the MET corrections in our ntuplewriter.py script were not running for MC in the latest CMSSW_7_4_X release. According to https://twiki.cern.ch/twiki/bin/view/CMS/MissingETUncertaintyPrescription#Instructions_for_7_4_X_X_12 they are not needed anymore. I removed the snippet from corMETFromMiniAOD.py completely because the corrections are supposed to be fully included in the default configuration. 

Could anyone else have a look into the twiki pages an confirm that this is correct?

### DIFF
--- a/core/python/ntuplewriter.py
+++ b/core/python/ntuplewriter.py
@@ -1,7 +1,7 @@
 import FWCore.ParameterSet.Config as cms
 
 
-useData = True
+useData = False
 use25ns = True #switch this flag to False when running on 50ns samples
 
 # minimum pt for the large-R jets (applies for all: vanilla CA8/CA15, cmstoptag, heptoptag). Also applied for the corresponding genjets.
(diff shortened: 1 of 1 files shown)


missirol:
based on the twikis (and the fact that the code doesn t run with this for 74X>12), i think this is the right thing to do. (i haven t check all the vars of the MET collections in the ntuple, though)

peiffer:
Thanks for cross-checking. The MET variables seems to be filled at least with some reasonable numbers.

marmeyer:
Hi Thomas, 

are the uncertainties of the MET still filled? 

Best, 

Mareike

peiffer:
Hi Mareike,

the uncertainties are filled for slimmedMETs and slimmedMETsNoHF. Only for the PUPPI collection the uncertainties are not working, but that has been the same in the previous version.

Cheers,

Thomas

marmeyer:
Ok, thanks for checking!

PR #279, opened 2015-11-21
added routine for jet-lepton cleaning by key-matching in common/

missirol:


peiffer:
Hi Marino,

thanks for implementing this function. In the normal JetLeptonCleaner we are also correcting the muonMultiplicity, muonEnergyFraction, electronMultiplicity and chargedEmEnergyFraction after the subtraction of the leptons. Is this not required in the key-based approach? It might be relevant for some jet-ID cuts.

Cheers,

Thomas

PR #288, opened 2015-11-24
added lumi files in new folder common/data

emanueleusai:


emanueleusai:
Well, I personally think that that the name of the root file should reflect which json was used (in terms of run numbers), rather than a generic name, for maximum transparency. 
However, I have nothing against implementing it in your way.
Maybe we can keep the root files there and add a softlink pointing to them with a generic name. 
Those who update the files should take care of updating the softlink as well.
What do you think?

peiffer:
The lumi file commit is fine, but why have you removed some files from SingleElectron_Run2015D-PromptReco-v4.xml?

dgonzal:
Thanks, I think that the having a link is the easiest way of having the latest recommended stuff with almost no additional work

HeinerTholen:
@peiffer those files are corrupt. I found they have an error connected to the source_candidate class. @missirol confirmed it as well, I think.

emanueleusai:
This is because @HeinerTholen and @missirol reported that these files were corrupted. These should have a minor impact on the luminosity (it's 3 files out of ca. 2800) so I just removed them. We'll have new ntuples productions before Moriond

emanueleusai:
Oh I see just now that Heiner was faster in answering...

PR #302, opened 2015-11-29
Fix in BTagScaleFactors; adding CSVv2 scale factor; minor updates

HeinerTholen:
BTagScaleFactors was producing NaN weights in some events through division by 0. The weights are now checked for being NaN, and if so, a runtime_error is thrown.


peiffer:
Hi Heiner,

could you also make a pull request for these commits to the master branch?

PR #310, opened 2015-12-02
muon scale factors and jet correction uncerts for subjets

HeinerTholen:


common/src/MCWeight.cxx
```diff
@@ -143,12 +143,102 @@ bool MCScaleVariation::process(Event & event){
...
+        eta_max_ > eta &&
+        pt_min_  < pt &&
+        pt_max_  > pt) {
+      int bin   = sf_hist_->FindFixBin(pt, eta);
+      float w   = sf_hist_->GetBinContent(bin);
+      float err = sf_hist_->GetBinError(bin);
+
+      weight      *= w;
+      weight_up   *= (w*(1+sys_error_percantage_) + err);
+      weight_down *= (w*(1-sys_error_percantage_) - err);
```
peiffer: Are the formulas correct here? And is sys_error_percantage really given in percent? If you say, that sys_error_percantage=1 gives you 1% of the systematic uncertainty, I would put sys_error_percantage to 100 to apply the full uncertainty (according to your comment in the header file). But then, these formulas here do not make sense for me. I would expect something like:

weight_up   *= w*(1 + sys_error_percantage_/100. * err);
weight_down *= w*(1 - sys_error_percantage_/100. * err);

HeinerTholen: Here's the description:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/MuonReferenceEffsRun2#Results_for_2015_data

In line 157, the percentage value from the user is divided by 100.
The stat error (err) is from GetBinError. Hence it has to be added to the weight, right?
Maybe the sys uncertainty has to be applied after the stat error, so it would look like this:
``weight_up *= (w + err)*(1+sys_error_percantage_);``
What do you think?

peiffer: Ah, okay, so sys_error_percantage is an ADDITIONAL uncertainty on the muon scale factor given in %. Then I would do 
weight_up *= w + err + w*sys_error_percantage_;
weight_down *= w - err - w*sys_error_percantage_;

HeinerTholen:
@peiffer: The uncertainties are now added in quadrature. Would you have a look at it?

peiffer:
Looks fine now. Thanks Heiner.

PR #311, opened 2015-12-02
muon scale factors and jet correction uncerts for subjets

HeinerTholen:


peiffer:
Hi Heiner,
could you please update this pull request with the changes from #310?

HeinerTholen:
It is already. The commit which is only in the RunII_25ns_v2 branch contains the sf data.

PR #323, opened 2015-12-18
QCD Samples

drberry85:
Hey Thomas,
I added all the QCD samples here at FNAL instead of the limited ones we had earlier.
Doug


common/src/TopJetIds.cxx
```diff
@@ -6,7 +6,7 @@ using namespace uhh2;
 
 bool CMSTopTag::operator()(const TopJet & topjet, const uhh2::Event &) const {
     auto subjets = topjet.subjets();
-    if(subjets.size() < 3) return false;
+    //if(subjets.size() < 3) return false;
```
peiffer: Hi Doug, I just realized (after accepting the pull request) that you have changed this line. What is the reason for this change? I guess this affects several users of the CMSTopTag routine.

peiffer: For the moment, I changed this line back to the original version with the subjets cut included.

drberry85: Thomas,

Good catch! I forgot I made that change. I shouldn't have committed it.

Doug

PR #330, opened 2016-01-14
Electron ID new vars

karavdin:
New variables for electrons related to HEEP cut

### DIFF
--- a/core/include/Electron.h
+++ b/core/include/Electron.h
@@ -1,13 +1,14 @@
 #pragma once
 
-#include "Particle.h"
+#include "RecParticle.h"
+//#include "Particle.h"
 #include "Tags.h"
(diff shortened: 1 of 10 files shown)


core/include/Electron.h
```diff
@@ -22,6 +23,8 @@ class Electron : public Particle {
     throw std::runtime_error("unknown Electron::tag '" + tagname + "'");
   }
 
+  typedef class source_candidate source_candidate;
```
missirol: this line was added only as a workaround for the previous 74X production in the RunII_25ns_v2 branch (see #274)
it was never needed in master; indeed the Muon class doesn't have it.
we should remove it. is it ok Thomas?

peiffer:
Hi Marino, thanks for spotting this. I will remove this line.

PR #333, opened 2016-01-17
Implement TopJetLepton cleaning and change JetCleaner and TopJetCleaner

dnowatsc:
# Commit 1
* Changed ntuplewriter so that lepton keys will be stored also for TopJets
* Changed  JetLeptonCleaner_by_KEYmatching so that it can also run over TopJets (tested it so that it doesn't break anybody's code)

# Commit 2
* Changed JetCleaner and TopJetCleaner classes so that they can run over arbitrary collections
* Classes in UHH2/common are changed accordingly so that the code compiles, however, users might have to change their private analysis code to account for this change. This simply requires adding ctx in the construcutor of the JetCleaner and TopJetCleaner classes

### DIFF
--- a/common/src/JetCorrections.cxx
+++ b/common/src/JetCorrections.cxx
@@ -429,7 +429,6 @@ JetLeptonCleaner_by_KEYmatching::JetLeptonCleaner_by_KEYmatching(uhh2::Context& 
 
     h_jets_ = ctx.get_handle<std::vector<Jet>>(jet_label);
     h_topjets_ = ctx.get_handle<std::vector<TopJet>>(jet_label);
-    label_ = jet_label;
     corrector = build_corrector(filenames);
     direction = 0;
     jec_uncertainty = corrector_uncertainty(ctx, filenames, direction) ;
@@ -461,6 +460,8 @@ bool JetLeptonCleaner_by_KEYmatching::do_cleaning(Jet & jet, uhh2::Event& event)
           muo_cand_p4.SetPy(muo_cand.py);
           muo_cand_p4.SetPz(muo_cand.pz);
           muo_cand_p4.SetE (muo_cand.E);
+
+          jet_p4_raw -= muo_cand_p4;
         }
       }
     }
@@ -484,6 +485,8 @@ bool JetLeptonCleaner_by_KEYmatching::do_cleaning(Jet & jet, uhh2::Event& event)
           ele_cand_p4.SetPy(ele_cand.py);
           ele_cand_p4.SetPz(ele_cand.pz);
           ele_cand_p4.SetE (ele_cand.E);
+
+          jet_p4_raw -= ele_cand_p4;
         }
       }
     }
--- a/common/src/JetCorrections.cxx
+++ b/common/src/JetCorrections.cxx
@@ -428,6 +428,8 @@ JetLeptonCleaner::~JetLeptonCleaner(){}
 JetLeptonCleaner_by_KEYmatching::JetLeptonCleaner_by_KEYmatching(uhh2::Context& ctx, const std::vector<std::string> & filenames, const std::string& jet_label){
 
     h_jets_ = ctx.get_handle<std::vector<Jet>>(jet_label);
+    h_topjets_ = ctx.get_handle<std::vector<TopJet>>(jet_label);
+    label_ = jet_label;
     corrector = build_corrector(filenames);
     direction = 0;
     jec_uncertainty = corrector_uncertainty(ctx, filenames, direction) ;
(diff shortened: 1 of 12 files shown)


common/src/JetCorrections.cxx
```diff
@@ -435,78 +437,98 @@ JetLeptonCleaner_by_KEYmatching::JetLeptonCleaner_by_KEYmatching(uhh2::Context&
...
+          correct_p4 = true;
 
-            jet_p4_raw -= muo_cand_p4;
-          }
+          LorentzVectorXYZE muo_cand_p4;
+          muo_cand_p4.SetPx(muo_cand.px);
+          muo_cand_p4.SetPy(muo_cand.py);
+          muo_cand_p4.SetPz(muo_cand.pz);
+          muo_cand_p4.SetE (muo_cand.E);
         }
```
missirol: where is the lepton-candidate p4 removed from jet_p4_raw in this new version? it seems jet_p4_raw is never modified in JetLeptonCleaner_by_KEYmatching::do_cleaning. have you tested the output of these modified routines?

dnowatsc:
You're completely right, somehow I must have deleted that line accidentally when I was cleaning up my code and removing some cout statements. I fixed and tested it, the module is now working again.
Thanks for the hint!

missirol:
general question: if top-jets are cleaned with the p4-subtraction method, shouldn't (one of the) subjets be cleaned as well? if not, after the cleaning, the subjets content is going to be inconsistent with the content of the corresponding top-jet.

common/src/JetCorrections.cxx
```diff
@@ -428,6 +428,8 @@ JetLeptonCleaner::~JetLeptonCleaner(){}
 JetLeptonCleaner_by_KEYmatching::JetLeptonCleaner_by_KEYmatching(uhh2::Context& ctx, const std::vector<std::string> & filenames, const std::string& jet_label){
 
     h_jets_ = ctx.get_handle<std::vector<Jet>>(jet_label);
+    h_topjets_ = ctx.get_handle<std::vector<TopJet>>(jet_label);
+    label_ = jet_label;
```
missirol: is this "label_" class member necessary? as far as i can see, it's not used anywhere

dnowatsc:
Concerning comment #1: In principle I agree but it all depends on what you want to do. For instance, I was cutting on the number of ak8 jets and the pt of the leading one (taken from the fat jet, not the subjets) and saw spikes in my muon pt distributions. Implementing TopJetLeptonCleaning (only on the fat jets, not the subjets) successfully removed these spikes since it removed all Ak8 jets that originated directly from a muon. If you're later using subjets e.g. to calculate the mass of the Ak8 jet and then cut on it, Ak8 jets originating from muons should be rejected anyway (since they probably don't even have subjets at all). Also we never used JetLepton cleaning for subjets of Ak8 jets, yet, we still used their subjets for all our mass calculations etc.

dnowatsc:
Concerning commen #2: yes, this is a left-over variable from when I was testing and debugging the code, thanks for pointing this out. Seems as if I'm not really good at code cleaning ;)

imarches:
Hi Dominik, didn't you have some time ago a plot showing that if you plot the pT of the leading top jet, calculated from subjets, the spike is not there? I think I remember something like that, but I'm not sure at the moment. This would confirm that, indeed, the substructure requirements already successfully kill the leptons.

dnowatsc:
@imarches: I'm don't think that I have such a plot right now and actually I'm not entirely sure whether this is true what you're saying. I remember seeing these spikes (in my muon pt spectrum) the first time when I was still using our patJetsHepTopTagCHSPacked_daughters collection, where all the ak8 jets are calculated from their subjets. But what I can say is that, if you use lepton cleaning with the fat jets (not the subjets) you can at least get rid of muons that were reconstructed as Ak8 jets (if you clean the TopJet collection afterwards) and I think in almost all cases, this is sufficient. However, I put a few lines in the documentation of this class to make users aware of the fact that subjets are not cleaned. Maybe in the future, it needs to be checked whether muons have a sizable effect on subjet-related quantities but right now, at least I don't have time to make this check myself.

dnowatsc:
Hi, can you please merge the PR? Or are there any issues left?

peiffer:
no, there are no issues left

peiffer:
Hi Dominik,
could you also make a pull request for these changes into the master branch?

missirol:
these updates don't introduce changes in the ntuple format, so they could/should also go to RunII_25ns_v2. right, @peiffer?

peiffer:
There is also a change to the NtupleWriter to store the electron and muon keys for TopJets.

missirol:
yes, but that doesn't change the ntuple format.. it's just filling a vector that was previously empty.
anyway that change could be taken out. i think it's better to have common/ in sync across these branches.

missirol:
Thomas, are we going to port this to RunII_25ns_v2 (without the change in core/plugins/NtupleWriter.cc, if necessary)?

peiffer:
Hi Marino, I have included it into RunII_25ns_v2, see #341

missirol:
thanks Thomas!

PR #334, opened 2016-01-19
systematic uncertainties in PUWeight

HeinerTholen:
I've restructured PUWeight a bit and added systematic uncertainties in the same fashion as for the BTagWeight class.

For MC, pileup_directory_50ns and pileup_directory_25ns are supported for backward-compatibility, but the new token should be pileup_directory.

Please review the code and let me know if it's fine. If so, I'll also create the PRs for the master and 76X branches.

### DIFF
--- a/common/include/MCWeight.h
+++ b/common/include/MCWeight.h
@@ -34,23 +34,28 @@ private:
 
 /** \brief Reweight MC based on simulated and actual pileup
  *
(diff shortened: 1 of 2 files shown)


peiffer:
Hi Heiner, the code looks fine. In principal, one could also write the routine in such a way that pileup_directory_data could be left out in the input xml card when doing the up or down variations. But I think, in this way it is also okay.

PR #354, opened 2016-02-03
Added root-files for MuonSFs and small changes to CommonModules

reimersa:
- Added root-files for Muon scale factors to common/data
- created a second argument for CommonModules::init to allow applying PU-systematics inside CommonModules


peiffer:
The change in CommonModules should also go into the master branch.

PR #362, opened 2016-02-09
Run ii 25ns v2

dgonzal:
Some additions to the code as described in the commit messages. should hopefully not corrupt any code


peiffer:
I think, this commit can also go into master and RunII_76X_v1.

PR #374, opened 2016-02-15
add option to configure the name of the xml parameter for MCBTagScaleFactor

emanueleusai:
This is needed in case btagging for different jet collections is used in the same analysis (e.g. AK4 and AK8). This way one can specify different efficiency files for the different jet collections and correct them separately.


peiffer:
Hi, I think this commit can also go into the master and the RunII_76X_v1 branches.

emanueleusai:
Will do!

PR #380, opened 2016-02-17
add possibility to set name postfix for stored btag weights

emanueleusai:
needed to avoid conflicts when applying btagging SF on more than one jet collection


peiffer:
Hi Emanuele,
I think this should also go into master and RunII_76X_v1.

PR #414, opened 2016-03-15
added tau uncertainties

mstoev:


peiffer:
Could you add some documentation to the header files? It would be nice to have the links to the twiki pages in the documentation to see where the hard-codes uncertainties are coming from. This commit should also go into the master branch.

PR #423, opened 2016-04-01
top pt reweighting is needed in several analysis

dgonzal:
code is comming from VLQ preselection repo, with small changes.


peiffer:
I think, this class can also go into the master branch.

PR #430, opened 2016-05-25
add AK8CHS,AK8PUPPI collections with low jet pT min

karavdin:


imarches:
Hi, unfortunately we need the collection as topjet collection. I suggest to close this PR and I can implemented as topjet collection myself

imarches:
Thanks for the clarification, Anastasia. Ok, we can have it also as normal jet collection. I have ameeting now. Thomas, if you want you can merge it, or I'll do it later.

PR #431, opened 2016-05-27
first JSON file for 2016 data

peiffer:


peiffer:
Lumi is 216022881.458/ub

PR #447, opened 2016-06-15
Adding xml of ST_t-channel

isando3:
Adding xml to the datasets for the ST_t-channel ntuples


peiffer:
Hi Daniel,

thanks for processing this sample. Please also correct the wiki page where still the aMC@NLO single top t-channel sample is listed, but this seems to be the Powheg anti-top sample. The Powheg top sample is still not there.

Thomas

PR #450, opened 2016-06-19
use subjet b tag SF along with AK4

emanueleusai:
- adding subjet b tag scale factors
- modification to MCBtagScaleFactor to allow to use AK4 and subjet btag SF at the same time


peiffer:
Hi Emanuele,

is this commit also useful for other productions? Then you should push it to the RunII_80X_v1 and master branch, too.

emanueleusai:
Yes, it is useful if one uses both subjet and AK4 btag and wants to use different scale factors for the two collections. I will propagate the change to the other branches after the preapproval

PR #484, opened 2016-07-21
Added MET correction for AK4CHS jets in JetCorrector()

reimersa:


common/include/JetCorrections.h
```diff
@@ -78,6 +78,7 @@ class JetCorrector: public uhh2::AnalysisModule {
     
     JetCorrectionUncertainty* jec_uncertainty;
     int direction = 0; // -1 = down, +1 = up, 0 = nominal
+    bool used_ak4chs = false, propagate_to_met = false;
```
peiffer: It would be sufficient ot define the variable used_ak4chs just locally in line 378.

PR #500, opened 2016-08-01
JEC propagation to MET

karavdin:
add slMETsCHS collection
add uncorrected MET variables
store L1 factor for JEC (needed for MET correction)

Result for JEC application and corresponding MET correction consistent with cms-jet/JMEReferenceTable results.


peiffer:
see #502

PR #560, opened 2016-12-04
new electron MVA IDs

peiffer:


peiffer:
see https://hypernews.cern.ch/HyperNews/CMS/get/egamma/1824.html for details

PR #588, opened 2017-01-11
make the module faster

dgonzal:
to calculate the area information for both hotvr and xcone makes the ntuplewrite 3-4 times slower! therefore we only calculate the area with reduced max eta of 4. For xcone only subjets have an area and for hotvr the ghost area is set to 0.02 instead of 0.01

### DIFF
--- a/core/src/UniversalJetCluster.cxx
+++ b/core/src/UniversalJetCluster.cxx
@@ -14,8 +14,13 @@ UniversalJetCluster::UniversalJetCluster(vector<PFParticle> *pfparticles)
     }
   ClusterHOTVR();
   //fastjet crashes with segmentation violation if not enough PFParticle are present!
-  if( pfparticles->size()>1)
+  if(pfparticles->size()>15)
     ClusterXCone33();
+
+  // to calculate the area information for both hotvr and xcone makes the ntuplewrite 3-4 times slower!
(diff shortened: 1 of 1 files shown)


core/src/UniversalJetCluster.cxx
```diff
@@ -154,10 +162,11 @@ void UniversalJetCluster::ClusterXCone33()
   // Run first clustering step (N=2, R=1.2) 
   vector<PseudoJet> fatjets;
   XConePlugin plugin_xcone(2, 1.2, 2.0);
-  double ghost_maxrap = 5.0;      // maxiumum y of ghost particles
+  double ghost_maxrap = 4.0;      // maxiumum y of ghost particles
   AreaDefinition area_def(active_area, GhostedAreaSpec(ghost_maxrap));
   JetDefinition jet_def_xcone(&plugin_xcone);
-  ClusterSequenceArea clust_seq_xcone(_psj, jet_def_xcone, area_def);
+  //ClusterSequenceArea clust_seq_xcone(_psj, jet_def_xcone, area_def);
+  ClusterSequence clust_seq_xcone(_psj, jet_def_xcone);
```
peiffer: Hi Daniel, but now you are switching off the area calculation at all.

PR #695, opened 2017-04-13
fix bug in tight JetID

karavdin:


karavdin:
Should go in the main branch too.

PR #698, opened 2017-04-15
fix bug with chsMET

karavdin:
the fix should go to master branch too.
By the way old line:
process.chs = cms.EDFilter("CandPtrSelector", src = cms.InputTag("packedPFCandidates"), cut = cms.string("fromPV")) 
does nothing, therefore not chs, but all packedPFCandidates were selected and used in all objects with 
src = cms.InputTag('chs') or src = 'chs', such as ak8CHSJetsSoftDrop, hepTopTagCHS, etc

### DIFF
--- a/core/python/ntuplewriter.py
+++ b/core/python/ntuplewriter.py
@@ -57,8 +57,10 @@ if isDebug:
 
 process.source = cms.Source("PoolSource",
   fileNames  = cms.untracked.vstring([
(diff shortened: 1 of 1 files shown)


peiffer:
Hi Anastasia, I think this pull request should only go to the master branch. We should keep the ntuplewriter.py script in the RunII_80X_v3 branch as it is for documentation of existing ntuples.

PR #735, opened 2017-07-09
Added Puppi TopJet Collection

drberry85:
Some basic book keeping to add a TopPuppiJetCollection to the SFrame config

### DIFF
--- a/core/include/Event.h
+++ b/core/include/Event.h
@@ -34,6 +34,7 @@ public:
   std::vector< Photon >* photons;
   std::vector< Jet >* jets;
   std::vector< TopJet >* topjets;
+  std::vector< TopJet >* toppuppijets;
   MET* met;
     
   GenInfo* genInfo;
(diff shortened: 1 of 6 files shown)


peiffer:
Hi Doug, do you need this additional jet collection in the central event or could you do this solely by the handle mechanism within your analysis module?

drberry85:
Thomas,

I need to keep the patJetsAk8PuppiJetsSoftDropPacked_daughters collection in the reduction step. And I thought that this was the best way to do that. Also, putting it in the central event, makes the matching very easy at the analysis level.

Doug

peiffer:
Hi Doug,
I submitted this commit for the RunII_80X_v3 branch. For the master branch we should stay with two default jet collections (normal jets and TopJets) to keep the code cleaner. In future analysis, we can certainly just take the Puppi collection as default TopJet collection and won't need two different TopJet collections in the event class.

drberry85:
Thomas,

Yeah, I think it's wise to put it on a separate branch. The reason for the two different collections is because of the use of both the slimmedJetsAK8_SoftDrop and Puppi collection.

Doug

PR #779, opened 2017-10-11
Enabled storing all variables for muons, electrons that were missing for 92X

reimersa:


karavdin:
It is just uncommenting of the old lines or something else is added to the script?

reimersa:
Mostly uncommenting of old lines. I needed to add some 
task.add(process.X), that's it.

PR #782, opened 2017-11-02
Run ii 80 x v3

drberry85:
Anastasia,

Another pull request for synchronization.

Doug


karavdin:
Hi Doug,
I have not seen such script before. What is it used for?

drberry85:
Anastasia,

This bsframe is how we run SFrame/UHH2/ZprimeSemiLeptonic on the condor queue at FNAL. It breaks the job up into many little jobs and ships them off to condor with a tarball of the CMSSW area. It can also automatically apply the systematics, but that part is a little old. So I wouldn't trust what it's doing for the PDF.

Doug

PR #789, opened 2017-11-23
Run ii 92 x v1

garbersc:
this pull contains the changes to the 92x ntuplewriter and FIXMEs at the spot that have to be changed to run 2016 MC with 2017 AnalysisModules (else triggerPrescales are looked for for MC and not found)

The files with a lot of changed were changed by my autoformatting...


garbersc:
now with the new xml files

PR #792, opened 2017-11-30
Updated filepath

MatthiesC:


raggleton:
The new filenames don't appear to exist either. E.g. 
```
ls: cannot access /pnfs/desy.de/cms/tier2//store/user/afroehli/Ntuples/Moriond17/BstarToTW_M-1200_LH_TuneCUETP8M1_13TeV-madgraph-pythia8/crab_BstarToTW_M-1200_LH_25ns/170109_111730/: No such file or directory
```
but there is
```
/pnfs/desy.de/cms/tier2//store/user/afroehli/Ntuples/Moriond17/BstarToTW_M-1200_LH_TuneCUETP8M1_13TeV-madgraph-pythia8/crab_BstarToTW_M-1200_LH_25ns/170131_103356/0000/*
```
Ditto for other random datasets I tried

MatthiesC:
Indeed, the filenames were still wrong. I sent another commit. It should work now.

PR #793, opened 2017-12-03
Fix fast jet linking [master]

raggleton:
Since they have deleted Thomas' AFS area, we now need to have another FastJet to link against.
We will use one in my AFS for now.

This PR updates the install script (which updates to tool files).

There will be separate PRs for the 80X and 92X branches.

For future, we will get user's to compile their own copy of FastJet to avoid this problem again.


karavdin:
install_CMSSW8026_withEGReg_UHH.sh is not really needed in the master branch

PR #784, opened 2017-11-13
Use scram to get inc/lib dirs for Makefiles

raggleton:
This uses the `scram tool tag <TOOL> <TAGNAME>` command to retrieve the locations of LHAPDF and FastJet `include`/`lib` dirs. This way the tool location only needs updating in `$CMSSW_BASE/config/toolbox/$SCRAM_ARCH/tools/selected/<TOOL>.xml`, and the Makefiles won't need updating.

This is necessary for both the move to 94X, and for remote deployments e.g. on gitlab

PR #795, opened 2017-12-03
Fix fast jet linking [RunII_92X_v1]

raggleton:
Since they have deleted Thomas' AFS area, we now need to have another FastJet to link against.
We will use one in my AFS for now.

This PR updates the install script (which updates to tool files), and backports #784 to ensure no hard-coded paths in Makefiles


karavdin:
Again install_CMSSW8026_withEGReg_UHH.sh is temporary script, not needed in 92X branch.

PR #802, opened 2017-12-14
Run ii 94 x v1

garbersc:
added to the .gitignore

added a warning to create-multi-dataset-xmlfiles.py to warn about potential missing double quotes that results in to much arguments

changed the basename extraction for files on dCache to match my setting, any easy idea on how to make it backwards competible for crab user?

added example gc-config file


raggleton:
I'm assuming that the new ntuplewriter config will be the only config for now... could you rename it "ntuplewriter.py" to replace the other one? Will avoid confusion, plus makes the diff useful

garbersc:
is updated

garbersc:
added first datasets

raggleton:
Thanks for renaming.

Couple of comments:

- Are the new datasets added in here produced using the ntuplewriter.py you've added in here?

- What are changes to the ntuplewriter? Is it just copy n paste from the 92X one? There's a lot of formatting changes that obscure content changes. 

- Also noticed the global tag is the 92X one, the 94X one is `94X_dataRun2_ReReco_EOY17_v2`

garbersc:
Yes they are. I updated th 92X to the changes I found in the ntuplewriter in this branch. I will rerun the processing with the correct tag.
Are there other changes were needed?

raggleton:
OK. I haven't checked through to ensure the jets etc are correct (although assuming not a whole load has changed). Let's go with it for now

reimersa:
Just going through the changes in ntuplewriter.py, did someone check that the lepton section, which apparently just has been commented, still works as intended? It seems to me that some variables like the effective area needed for calculation electron isolations are switched off.

raggleton:
I haven't. I assume they were switched off in 92X for a specific reason. But it will be reviewed for 94X (just haven't got round to it yet).

reimersa:
I switched them back on for 92X and made them work, but only in ntuplewriter.py, not _92X.py. The lepton part was reset by renaming the script. This part is needed for filling non-standard lepton variables, it should be restored at some point.

----- Original Message -----

PR #803, opened 2017-12-14
updated global tag

garbersc:


raggleton:
Ah while you're at it, can you change the MC one to `94X_mc2017_realistic_v10`? thanks!

PR #805, opened 2017-12-15
overlooked the smearing

garbersc:
sorry

### DIFF
--- a/common/src/JetCorrections.cxx
+++ b/common/src/JetCorrections.cxx
@@ -1172,6 +1172,29 @@ const JERSmearing::SFtype1 JERSmearing::SF_13TeV_2016 = {
   {{5.0, 1.160, 1.189, 1.131}},
 };
 
+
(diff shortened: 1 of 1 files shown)


common/src/JetCorrections.cxx
```diff
@@ -1172,6 +1172,29 @@ const JERSmearing::SFtype1 JERSmearing::SF_13TeV_2016 = {
   {{5.0, 1.160, 1.189, 1.131}},
 };
 
+
+// https://twiki.cern.ch/twiki/bin/view/CMS/JetResolution [13TeV JER measurement]
+//New JER files from Marek (8 Aug 2017)
+const JERSmearing::SFtype1 JERSmearing::SF_13TeV_2016_03Feb2017 = {
```
raggleton: Maybe odd question - presumably they are called "03Feb2017" because they use that dataset, and has nothing to do with when Marek made them (8 Aug 2017)...?

garbersc: yes, its the name of the 2016 rereco datasets

PR #811, opened 2018-01-08
[94X] Update Ntuple electrons with new VID

raggleton:
This stores electrons with the preliminary VID from EGamma: https://hypernews.cern.ch/HyperNews/CMS/get/egamma/2006.html

At the moment, they are only the cut-based IDs. The appropriate tags have been added to the `Electron` class, replacing the Summer16 ones (since I assume they're not useful at this point?).
Note that I **haven't** update the `ElectronIds.[h|cxx]` files, as I thought just using the tag in an analysis would be easier, and less likely to lead to transcription errors. 

Also updated install script.

### DIFF
--- a/core/include/Electron.h
+++ b/core/include/Electron.h
@@ -15,20 +15,20 @@ class Electron : public RecParticle {
     twodcut_dRmin,
     twodcut_pTrel,
     heepElectronID_HEEPV60,		
-    cutBasedElectronID_Summer16_80X_V1_veto,
-    cutBasedElectronID_Summer16_80X_V1_loose,
-    cutBasedElectronID_Summer16_80X_V1_medium,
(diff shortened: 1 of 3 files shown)


raggleton:
I should add that when running, it will spit out:

```
%MSG-w IdInformation:  VersionedGsfElectronIdProducer:egmGsfElectronIDs@ctor  08-Jan-2018 19:02:26 CET pre-events
Instantiated ID: cutBasedElectronID-Fall17-94X-V1-Preliminary-loose
with MD5 hash:
This ID is not POG approved and likely under development!!!
Please make sure to report your progress with this ID at the next relevant POG meeting.
...
```

This is to be expected as EGamma marked it "not approved".

PR #813, opened 2018-01-09
add TopJets, minor cleaning

karavdin:


core/python/ntuplewriter.py
```diff
@@ -658,6 +654,9 @@ def add_fatjets_subjets(process, fatjets_name, groomed_jets_name, jetcorr_label=
                  elSource=cms.InputTag('slimmedElectrons'),
                  getJetMCFlavour=(not useData)
                  )
+#manually override parton & genjet matching even though we set getJetMCFlavour false...
+#ToDo: figure out more beautiful solution
```
raggleton: Hmm I'm having a look to see if this as been solved - someone has already mentioned it: https://github.com/cms-sw/cmssw/issues/20057

raggleton:
So having stared at code, I don't think there's a nice way to turn off the gen* parts in `addJetCollection()` annoyingly. I'll merge this, then make a PR to tidy up this manual patching of `addJetProducer`

PR #824, opened 2018-01-18
[94X] Use Puppi JEC for Puppi jets

raggleton:
This changes the AK8Puppi jets (& subjets) to use the PUPPI JEC, not the CHS ones. Since this also affects the "normal" AK8PUPPI jets, is it OK for JEC work as well?

(@abenecke pointed out that a change in master & 92X ntuplewriter.py didn't get propagated through.)


karavdin:
@raggleton, @abenecke 
The change is ok, but please keep in mind that we usually use some dummy JEC during the production and jets must be re-corrected in each analysis

PR #825, opened 2018-01-19
[94X] Update HEEP & MVA Electron IDs

raggleton:
Update electron IDs to store HEEPv7.0, and the latest MVA IDs.
See https://hypernews.cern.ch/HyperNews/CMS/get/egamma/2013.html

**NOTE** This requires doing an extra bit of setup, see changes in `install.sh`. You should also `make clean && make` to ensure all the changes in `uhh::Electron` are picked up properly.

For HEEP, we only store the pass/fail bool.
For MVA IDs, we store both the MVA values and the pass/fail bools.
The changes to the MVA bits are more extensive as now there are 2 MVA IDs, iso & non-iso, so `uhh::Electron` methods have been updated, as have IDs.
Since the wp80 and wp90 are now not just simple cut values but a formula, I haven't included them in the `ElectronIds.cxx`.
If they ever change, we can redo the calculation using the stored MVA values.
The loose working point is a simple cut value, so I've added it in.

### DIFF
--- a/common/include/ElectronIds.h
+++ b/common/include/ElectronIds.h
@@ -255,59 +255,44 @@ namespace ElectronID {
 // Electron MVA IDs
 // REF https://twiki.cern.ch/twiki/bin/view/CMS/MultivariateElectronIdentificationRun2
 
-// --- General Purpose MVA ID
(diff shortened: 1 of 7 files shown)


karavdin:
Hi Robin,
The changes might be useful for the master branch too.

PR #826, opened 2018-01-22
Run ii 94 x v1

garbersc:
added first 94x MC xml


karavdin:
@garbersc,
I believe TEMPLATE_CONFIG* files are not needed. Could you remove them, please?

PR #831, opened 2018-01-23
Refreshed muons to store additional information

reimersa:
- Enabled muon variables that were still missing
- Now we store all pre-computed IDs and isolations provided by MUO as bits.
- Changed to a more flexible way to access muon ID, isolation. Note that this requires most analyses to tweak their code a bit
- Now we store additional information about the muon type (primary, punchthrough, from quark decays, ...)

### DIFF
--- a/common/include/MuonIds.h
+++ b/common/include/MuonIds.h
@@ -5,25 +5,15 @@
 
 // see also ElectronIds.h for general comments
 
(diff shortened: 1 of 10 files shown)


raggleton:
Looks good, thanks for adding in all the vars & selectors.

PR #835, opened 2018-01-26
[94X] Remove our CHS MET producer, store the official one from MiniAOD

raggleton:
This retires our slMETsCHS collection, in favour of storing the CHS "correction" like as is done in MiniAOD.
The CHS MET is now accessible from `slimmedMETs`, using the `corrPx_CHS()` and `corrPy_CHS()` methods.

The use of "corr" is in keeping with the names from pat::MET: https://github.com/cms-sw/cmssw/blob/CMSSW_9_4_1/DataFormats/PatCandidates/interface/MET.h#L151-L168
Although it should be pointed out that the MET itself **doesn't** have any Type-1 corrections.

@karavdin  / anyone that uses any of these MET corrections: is storing px and py useful? Or is storing pt/phi better? Or the whole LorentzVector?

Unfortunately this saves very little processing time, the previous modules were very efficient:
```
TimeReport   0.000789     0.000789     0.000789  pfMet
TimeReport   0.000019     0.000019     0.000019  slMETsCHS
```

I also compared the values of our slMETsCHS vs the value in MiniAOD, and they were practically identical (pt matched within ~0.005 GeV or better, difference due to lost info when we use packedPfCandidates).

### DIFF
--- a/core/include/MET.h
+++ b/core/include/MET.h
@@ -33,6 +33,8 @@ public:
       m_shiftedPy_TauEnDown = 0;
       m_shiftedPy_MuonEnDown = 0;
       m_shiftedPy_MuonEnUp = 0;
+      m_corrPx_CHS = 0;
+      m_corrPy_CHS = 0;
       /* m_corr_x = 0; */
       /* m_corr_y = 0; */
       /* //  m_corr_SumEt = 0; */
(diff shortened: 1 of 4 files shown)


karavdin:
Hi @raggleton,
Storing the LorentzVector directly will be useful, but one can build it from px and py. The name is misleading. Can we put a comment about it in the code, please?

PR #837, opened 2018-01-30
[94X] Remove CA15 pruned mass

raggleton:
Remove the CA15 pruned mass (for HepTopTagger collection). I don't think it will be used (we already store the mass of the groomed fat jet the top tagger produces), and it is remarkably slow for one variable (~0.15s / event).


raggleton:
Eurgh screwed up commits, will redo this

PR #842, opened 2018-01-31
[94X] Rename CHS MET variable

raggleton:
Old name implied it was corrected, definitely isn't. More suitable @karavdin ?


karavdin:
Great! Thank you :)

PR #848, opened 2018-02-01
Run ii 94 x v1

garbersc:
added 17Nov2017 L123 JERC pathes for runs B-F and MC

changes to ntuplewriter are just autoformatting


karavdin:
Hi @garbersc,
Could you remove ntuplewriter.py from the commit, please? I don't see any important changes, but it will destroy formating of the current version.

garbersc:
ntuplewriter back to old format

PR #850, opened 2018-02-02
configs for production without HOTVR and XCONE

karavdin:
samples for JEC related studies produced without HOTVR and XCONE  due to lack of tims


raggleton:
As a future thought, I think we should somehow make the `process` object from `ntuplewriter.py` import-able. This would require some sort of generating function in `ntuplewriter.py` so you can set `useData`. Then you can do e.g.

```
from ntuplewriter import create_process
myProcess = create_process(useData=True)
myProcess.MyNtuple.doHOTVR = cms.bool(False)
myProcess.MyNtuple.doXCone = cms.bool(False)
```

This would also make testing over a variety of inputs/settings easier

PR #852, opened 2018-02-02
[94X] Ntuplewriter with multithreading

raggleton:
This allows the ntuplewriter to run with multiple threads, significantly speeding up runtime. In local tests running over 500 MC events, I went from ~4s/evt (1 thread) to ~0.9s/evt (4 threads) and ~0.5s/evt (8 threads). I haven't checked memory usage yet though.

To run `cmsRun` in multithreaded mode, do e.g. `cmsRun -n 4 ntuplewriter.py` where 4 is the number of threads.
You can pick any number, but it shouldn't be more than the number of cores available, and as you use more you get diminishing returns.

**NB** To use multithreading, we now have to use the CMS versions of fastjet and fastjet-contrib, which have thread-safe additions. If you want to run multithreads, you will have to re-install them, see `install.sh`. (The easiest way is to delete the contents of your `fastjet-install` dir, follow the install instructions, and then redo the `scram setup fastjet && scram setup fastjet-contrib && scram setup fastjet-contrib-archive && scram b clean && scram b -j12`, and then re-compile UHH.

I updated all our ED modules to the new stream/global/one types, and replaced any `cout`s or `runtime_error` with `edm::LogInfo/Warning` and `cms::Exception`, respectively.

I also added in a script I use to compare all leaves in all branches of two TTrees, `compareTrees.py`. It prints plots of every variable, and tells you if two plots differ (not the most sophisticated way though).
Useful for testing 1 core vs multicore output. Noticeably, I found the HOTVR and XCONE jet areas would differ slightly. This is due to having a non-deterministic random number seed. I copied the method used in the CMSSW `VirtualJetProducer` to set the seed deterministically based on event and run numbers, and the jet area now agrees for 1-core and multi-core tests.

I also spotted a few uninitialised variables that needed setting.

**NB2** I tried a very small number of jobs on CRAB, and it seems to work. However, I don't know what will happen when it scales to > 10K jobs, you may get stuck in a queue for slots with >1 CPU. Or there may be some new race-condition I haven't seen yet.
To run on CRAB you have to set in `ntuplewriter.py`:
```
process.options.numberOfThreads = cms.untracked.uint32(4)
```
and in your crab script:
```
config.JobType.numCores = 4
```
CRAB only accepts 1, 4, or 8 cores, and the number in `ntuplewriter.py` must be the same.

### DIFF
--- a/core/include/GenTopJet.h
+++ b/core/include/GenTopJet.h
@@ -5,6 +5,10 @@
 class GenTopJet : public Particle {
 public:
 
+  GenTopJet() {
+    m_tau1 = m_tau2 = m_tau3 = m_chf = m_cef = m_nhf = m_nef = -1.0;
+  }
+
   const std::vector<Particle> & subjets() const{return m_subjets;}
   void add_subjet(const Particle & p){m_subjets.push_back(p);}
   float tau1() const{return m_tau1;}
--- a/core/include/JetBTagInfo.h
+++ b/core/include/JetBTagInfo.h
@@ -38,11 +38,17 @@ public:
     m_FlightDistance3dVal.clear();
     m_FlightDistance3dSig.clear();
     m_VertexJetDeltaR.clear();
+    m_JetNSecondaryVertices = 0;
     m_VertexNTracks.clear();
     m_SecondaryVertex.clear();
     m_VertexChi2.clear();                                
     m_VertexNdof.clear();                    
     m_VertexNormalizedChi2.clear();
+
+    m_VertexCategoryJTC = 0;
+    m_VertexMassJTC = 0;
+    m_VertexEnergyRatioJTC = 0;
+    m_TrackSip3dSigAboveCharmJTC = 0;
   }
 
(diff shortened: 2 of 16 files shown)


raggleton:
Just did a quick memory test. For TTbar MC with 500 evts, with 1 thread it peaked at ~1500MB, whilst for 4 threads it peaked at ~1900MB. That's not great, given that (a) it will likely grow far more with a few K events, (b) the CRAB default is 2GB (although in the past we requested 2.5GB).

raggleton:
Tried submitting crab jobs for Single Mu Run C ~24 hours ago. Requested 4 cores, 2.5GB RAM, 20K events/job (~9K jobs in total). Don't know if unusually slow, but nearly 24hr later and its only done ~1/4 of them, with about half still pending. In comparison Torben did Single Mu Run D single threaded (10K events/job) and his was ~95% done in 24 hours. So either this dataset is being slow, or asking for 4 cores makes it a lot harder to find a free job slot.

PR #858, opened 2018-02-06
[94X]  fix bug to store CHS MET x and y components and add correct_MET for CHS MET

karavdin:
fix bug to store CHS MET x and y components and not y for both.
correct_met for CHS MET and slimmedMETs are decoupled now. To use it for CHS MET user should call correct_met(event, true). This will set MET to TypeI corrected CHS MET. User also should set L1RC corrector to work with CHS MET. 
The standard call  correct_met(event) will use input collection specified by user (as before, so no change visible for most of users).
Also some cleaning is done to remove not necessary any more flags, etc


core/plugins/NtupleWriter.cc
```diff
@@ -1124,7 +1124,7 @@ bool NtupleWriter::filter(edm::Event& iEvent, const edm::EventSetup& iSetup) {
                met[j].set_shiftedPy_TauEnDown(pat_met.shiftedPy(pat::MET::METUncertainty::TauEnDown, pat::MET::METCorrectionLevel::Type1));
                met[j].set_shiftedPy_MuonEnDown(pat_met.shiftedPy(pat::MET::METUncertainty::MuonEnDown, pat::MET::METCorrectionLevel::Type1));
                met[j].set_shiftedPy_MuonEnUp(pat_met.shiftedPy(pat::MET::METUncertainty::MuonEnUp, pat::MET::METCorrectionLevel::Type1));
-               met[j].set_rawCHS_px(pat_met.corPy(pat::MET::METCorrectionLevel::RawChs));
+               met[j].set_rawCHS_px(pat_met.corPx(pat::MET::METCorrectionLevel::RawChs));
```
raggleton: Ack, sorry about that :(

raggleton: In future ntuple iterations, perhaps we should move to pt and phi, to (a) avoid typos as diff lengths, (b) more useful to analysist

PR #860, opened 2018-02-07
[94X] Add in ReReco SingleMu RunC XML

raggleton:
With HOTVR & XCONE. Tested all are valid.

Name is in line with the JetHT XMLs, but dunno if we want to prepend "DATA_"?


karavdin:
Hi Robin,
Yes, we usually add DATA_ to make search in ~1000 xml files easy :)

PR #866, opened 2018-02-08
Run ii 94 x v1

garbersc:
added Fall17_17Nov2017_V4 JERC paths


common/src/JetCorrections.cxx
```diff
@@ -1078,9 +1143,12 @@ bool JetLeptonCleaner::process(uhh2::Event & event){
         }
     }
     if(event.electrons){
+  std::cout<<"event.electrons true"<<std::endl;
```
raggleton: Can you remove the couts please

garbersc: oh sorry

PR #875, opened 2018-02-14
Added TTbar semileptonic MC

reimersa:


raggleton:
Hi, are this with or without XCONE & HOTVR? (sorry can't see the filenames, diff too large). If so could you put them in the RunII_94X_v1_woHOTVR_woXCONE directory instead so it's all consistent? Thanks

reimersa:
w/o XCone, HOTVR. I moved the file.

raggleton:
Thanks!

PR #876, opened 2018-02-14
Add TTTo2LL and DY background samples

froehlia:


raggleton:
Hi, are this with or without XCONE & HOTVR? (sorry can't see the filenames, diff too large). If without could you put them in the RunII_94X_v1_woHOTVR_woXCONE directory instead so it's all consistent? Thanks

froehlia:
Hi, these are without XCONE & HOTVR. I moved them to the correct directory.

raggleton:
Thanks!

PR #879, opened 2018-02-14
Updated example/Makefile

MatthiesC:
With this update, the create-new-analysis compiles again

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,5 +1,5 @@
 LIBRARY := SUHH2examples
-USERLDFLAGS := -lSUHH2core -lSUHH2common -lGenVector
+USERLDFLAGS := -lSUHH2core -lSUHH2common -lGenVector -lSUHH2JetMETObjects -L/cvmfs/cms.cern.ch/slc6_amd64_gcc530/external/lhapdf/6.1.6-ikhhed/lib -lLHAPDF
 # enable par creation; this is necessary for all packages containing AnalysisModules
 # to be loaded from by AnalysisModuleRunner.
 PAR := 1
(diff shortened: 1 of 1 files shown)


raggleton:
What was the error before? During compilation or whilst running `sframe_main`? I just tried create-new-analysis without the LHAPDF libs and it compiled OK.

reimersa:
The LHAPDF libs are needed when evaluating PDF uncertainties for signal samples without corresponding weights, e.g. pure pythia samples. Everything compiles in the beginning but eventually those flags might be needed and digging out the problem then is much more work. Also, the JetMET libs were missing such that some classes in https://github.com/UHH2/UHH2/tree/master/JetMETObjects could not be used.

raggleton:
OK good to know, never tried the PDF uncertainties myself

MatthiesC:
Thanks for the explanation, Arne

raggleton:
So this can be ported to 94X, master, etc it would be better to avoid hard-coding the LHAPDF lib directory, and instead just ask scram for it. Could you add in

```
LHAPDFLIB=$(shell scram tool tag LHAPDF LIBDIR)
```
before the `USERLDFLAGS`, and then replace

```
-L/cvmfs/cms.cern.ch/slc6_amd64_gcc530/external/lhapdf/6.1.6-ikhhed/lib
```
with

```
-L${LHAPDFLIB}
```

And `make clean && make` again to check it works. Thanks

examples/Makefile
```diff
@@ -1,5 +1,6 @@
 LIBRARY := SUHH2examples
-USERLDFLAGS := -lSUHH2core -lSUHH2common -lGenVector
+HAPDFLIB=$(shell scram tool tag LHAPDF LIBDIR)
```
raggleton: Missing the `L` here

PR #880, opened 2018-02-15
added line with weighted number of events

reimersa:


raggleton:
Does anything in SFrame use this tag? Or is it just a record from running the readaMCatNloEntries.py script?

reimersa:
It's just the record from readaMCatNloEntries.py

PR #884, opened 2018-02-19
More robust b/W daughter identification in TTbarGen

MatthiesC:
New attempt without re-indentation: Added workaround in TTbarGen for the rare issue that errors occur due to tops having no b/W daughters. Now, b/W with top mothers are looked for if the initial daughter search fails. The fact, that it must be W+ from top and W- from anti-top, is taken into account. Same for the bottoms. Should probably be in the master as well. Runtime errors are de-commented again.

### DIFF
--- a/common/src/TTbarGen.cxx
+++ b/common/src/TTbarGen.cxx
@@ -15,24 +15,54 @@ TTbarGen::TTbarGen(const vector<GenParticle> & genparticles, bool throw_on_failu
             auto w = genp.daughter(&genparticles, 1);
             auto b = genp.daughter(&genparticles, 2);
             if(!w || !b){
-	      //if(throw_on_failure) throw runtime_error("TTbarGen: top has not ==2 daughters");
-                continue;
+	      if(throw_on_failure) throw runtime_error("TTbarGen: top has not ==2 daughters");
+              return;
             }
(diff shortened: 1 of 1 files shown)


raggleton:
Hi, sorry it's taken me a while to look at this, at a school at the moment. Do you have any example event(s) that have top decays with no b/W daughters I could run over? Curious to see what's going on in them

MatthiesC:
For example files, see here:
/afs/desy.de/user/m/matthies/split_script_TTbar.sh.o5242218.57
It's the output of one of the failing split scripts when I try to run the original TTbarGen. The corresponding error output file says that no W daughter was found.
Here, I modified TTbarGen in such a way that it also prints the genparticles list for the failing event. See end of output file.
It looks like the antitop (ind 3) is mother to 6 particles instead of 2 (b, W, and four pairly produced electrons/positrons). From these six particles, just two are declared to be "the" top daughters. In this case, two positrons --> no b/W daughter found

raggleton:
Can you put the file in your `public` directory on AFS? Otherwise I cant access it

MatthiesC:
Done

raggleton:
OK, I see what you mean now. That's a bit odd... I guess you really have to rely on the particle's mother as being correct, rather than the daughter indices of the parent.

PR #885, opened 2018-02-19
update on muon scale factors

tordreyer:
Muon ID scale factors are just provided as .json files and not as .root files. I added a script that converts the json file into a root file.

Root files for Muon ID and Trigger scale factors are added to common/data/.

I changed the name of the histogram that we read from the file from pt_abseta to abseta_pt because only abseta_pt supported. It is just a change of the axis.


karavdin:
Hi @tordreyer,
Does "MuonTrigger_EfficienciesAndSF_RunBtoF_Nov17Nov2017.root" contains SFs for set of triggers or for just one trigger?

tordreyer:
It contains SFs for a set of triggers. It is taken from: https://twiki.cern.ch/twiki/bin/view/CMS/MuonReferenceEffs2017

PR #888, opened 2018-02-21
Run ii 94 x v1

garbersc:
added reprocessed RunE xml 

and all use of genparticles is in try-catch structures that supress error due to missing genparticle information in files used by BaconJets
keep it or set it back?


karavdin:
Might be useful for the future too. Thanks!

raggleton:
What was the issue with the genInfo stuff? Does it just randomly fail?

garbersc:
no, we had a ntuple processing run of the MC were the gen information was not saved due to a misset flag

core/src/EventHelper.cxx
```diff
@@ -37,6 +37,8 @@ EventHelper::EventHelper(uhh2::Context & ctx_): ctx(ctx_), event(0), pvs(false),
 #define IMPL_SETUP(name, type) void EventHelper::setup_##name (const std::string & bname) {\
     if(bname.empty()) return; \
     name = true; \
+    std::cout<<"going to declare "<<bname<<std::endl; \
```
raggleton: Please remove

karavdin:
Hi @garbersc,
Something is broken in this PR. After the switch to it I can't run MC samples any more. However it was running fine on the same sample before the switch.
The error:
Event: error for member of type 'GenInfo' at index 18446744073709551615: get: index out of bounds, i.e. the handle used is invalid
Appears  for following line:
double pt_hat = ev.genInfo->binningValues()[0]

Did you test it on MC?

core/src/EventHelper.cxx
```diff
@@ -37,6 +37,8 @@ EventHelper::EventHelper(uhh2::Context & ctx_): ctx(ctx_), event(0), pvs(false),
...
+    if(bname == "genInfo") return; \
```
raggleton: @karavdin I'd guess this is why it stopped working on MC for you?

PR #892, opened 2018-02-26
Run ii 94 x v1

garbersc:
added V6 V7 JEC and new RunE xml with 15 of 9001 jobs missing


karavdin:
Fal17_17Nov V5,V6 and V7 does not exist for MC. Please remove it from the JetCorrector

PR #897, opened 2018-04-16
Run ii 94 x v1

garbersc:
updated the tight IDs according to https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetID13TeVRun2017

### DIFF
--- a/common/src/JetIds.cxx
+++ b/common/src/JetIds.cxx
@@ -128,5 +128,5 @@ bool JetPFID::tightLepVetoID(const Jet & jet) const{
   if(jet.muonEnergyFraction() <0.8
      &&jet.chargedEmEnergyFraction()<0.80)
     return true;
-  return false
+  return false;
 }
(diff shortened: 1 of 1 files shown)


common/src/JetIds.cxx
```diff
@@ -92,20 +93,40 @@ bool JetPFID::looseID(const Jet & jet) const{
...
+    
+    if(fabs(jet.eta())>=2.4)
+      return true;
+      
+    if(jet.chargedHadronEnergyFraction()>0
+       && jet.chargedMultiplicity()>0)
+      return true;   
+  }
+  else if(fabs(jet.eta())>2.7 && fabs(jet.eta())<=3
+	  &&jet.neutralEmEnergyFraction()<0.90
```
raggleton: This should be 0.99 according to the twiki page

raggleton:
For the loose ID, since it's no longer an ID in 2017, should we (a) just get rid of it, or (b) add a depreciation printout to the user so they know that the official JetID is either Tight or TightLepVeto?
My worry is that someone transferring their analysis to 2017 will use it and no know that its doing ~nothing.

common/src/JetIds.cxx
garbersc: you are right, I corrected this

raggleton:
OK I'll merge this, and then do a separate PR for removing the loose ID

PR #907, opened 2018-05-08
Updated cut based electron IDs for 2017

reimersa:


raggleton:
If I understand correctly, these should give the same ID results as the stored tags in our `Electron` class (e.g `cutBasedElectronID_Fall17_94X_V1_loose`). Would it be possible to do a quick check to ensure that the status from your changes & the stored statuses agree?

PR #909, opened 2018-05-11
updated JEC and MET correction

garbersc:
re-enabled V6 MC JEC
added option for thresholds in the MET correction

### DIFF
--- a/common/src/JetCorrections.cxx
+++ b/common/src/JetCorrections.cxx
@@ -810,12 +810,12 @@ const std::vector<std::string> JERFiles::Fall17_17Nov2017_V5_F_L1RC_AK4PFchs_DAT
 
 
 //2017-------------------------- 17Nov2017  V6
-// const std::vector<std::string> JERFiles::Fall17_17Nov2017_V6_L123_AK4PFchs_MC = {
-//   "JECDatabase/textFiles/Fall17_17Nov2017_V6_MC/Fall17_17Nov2017_V6_MC_L1FastJet_AK4PFchs.txt",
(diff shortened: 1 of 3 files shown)


common/src/JetCorrections.cxx
```diff
@@ -1072,12 +1072,15 @@ std::unique_ptr<FactorizedJetCorrector> build_corrector(const std::vector<std::s
...
   //NB: jet with substracted muon Pt should be used
-  void correct_MET(const Event & event, const bool & do_L1corr){
+  void correct_MET(const Event & event, const bool & do_L1corr, double pt_thresh, double eta_thresh_low, double eta_thresh_high){
     //we start from raw MET
     LorentzVector metv4= event.met->uncorr_v4();
     for(auto & jet : *event.jets){
       //thresholds on the corrected jets: pt > 15, EM fraction < 0.9
-      if(jet.v4().Pt() > 15 && (jet.neutralEmEnergyFraction()+jet.chargedEmEnergyFraction())<0.9){
+      bool to_be_corrected = jet.v4().Pt() > 15.;
+      to_be_corrected = to_be_corrected && ( fabs(jet.v4().Eta())<eta_thresh_low || fabs(jet.v4().Eta())>eta_thresh_high || jet.v4().Pt() > pt_thresh );
```
raggleton: Are these meant to be all ORs? e.g. if you had `eta_thres_low = 1`, `eta_thres_high=3`, `pt_thres=20`, a jet with eta = 0.1 and pt = 18 would still pass selection. Would have thought you want to pass the eta AND pt requirements?

raggleton: Also I assume the intention is to select jets _outside_ or a certain eta window, and not inside?

garbersc: the intention is NOT to correct jets in a eta-window below a certain pt-threshhold

PR #912, opened 2018-05-16
[80X_v4] Fix DasQuery.py

raggleton:
Copy of #902 for the 80X_v4 branch


raggleton:
This should be the only thing done in v3 that wasn't done in v4, after this I can create a new branch from v4 for all the XCONE/HOTVR changes

PR #916, opened 2018-05-17
[80Xv5] Update XCone/HOTVR

raggleton:
Updates for HOTVR/XCone:

**Physics changes**:

- Ntuple now has XCone & HOTVR with both CHS & PUPPI, dropping the pfCandidates ones (i.e. what we used to have)
- Move to `PseudoXConePlugin` instead of normal `XConePlugin` for speed gains (both reco & gen level). This saves us ~0.3s/evt/producer, so about 0.6s/evt overall. For reco XConeProducer this is configured by a bool to allow easy testing/changing. For gen level its hard-coded.
- Both producers now filter out input particles with pT= 0.
- Gen level XCone & HOTVR now use genParticles **without neutrinos** - previously had included them. (NB this doesn't affect runtime)

**Code changes**:

- Move clusterers  into own producers and out of UnivesalClusterer
- NtupleWriter can now store any number of XCone/HOTVR collections (reco only)
- Made use of `puppi.useExistingWeights` to shave off time (about 0.2s/evt)
- Had to do some internal changing to allow modules to consume output from `CandPtrSelector` (i.e. take `reco::Candidate`)
- Some other niceties like fixing the const warnings in `GenTopJet` when compiling, init instance vars, deterministic RNG seed for clusterers

For comparison, here are the runtimes with full XCone, recalculating puppi weights:

```
TimeReport      event loop Real/event = 2.931695

TimeReport   0.072718     0.072718     0.072718  MyNtuple (includes gen level XCone, HOTVR)
TimeReport   0.161843     0.161843     0.161843  hotvrCHS
TimeReport   0.075718     0.075718     0.075718  hotvrPuppi
TimeReport   0.204478     0.204478     0.204478  puppi
TimeReport   0.437445     0.437445     0.437445  xconeCHS
TimeReport   0.360391     0.360391     0.360391  xconePuppi
```

and after all changes above:

```
TimeReport      event loop Real/event = 2.012205

TimeReport   0.021691     0.021691     0.021691  MyNtuple
TimeReport   0.005801     0.005801     0.005801  puppi
TimeReport   0.102966     0.102966     0.102966  xconeCHS
TimeReport   0.091758     0.091758     0.091758  xconePuppi
```
so all the changes shave off 1s/evt.

TBH I don't see how we can get under 2s/evt without dropping more producers/collections.

One thing that is needed is a mem check to make sure none of the new modules introduced are creating havoc.

For reference, here was the timings with 80Xv4:

```
TimeReport      event loop Real/event = 2.303859

TimeReport   0.587225     0.587225     0.587225  MyNtuple (has all the XCone, HOTVR stuff inside it)
TimeReport   0.201786     0.201786     0.201786  puppi
```

so overall we gain a little speedup and have more useful collections 😄

### DIFF
--- a/core/include/GenTopJet.h
+++ b/core/include/GenTopJet.h
@@ -11,19 +11,19 @@ public:
 
   const std::vector<Particle> & subjets() const{return m_subjets;}
   void add_subjet(const Particle & p){m_subjets.push_back(p);}
-  const float tau1() const{return m_tau1;}
+  float tau1() const{return m_tau1;}
   void  set_tau1(float tau1){m_tau1=tau1;}
-  const float tau2() const{return m_tau2;}
+  float tau2() const{return m_tau2;}
   void  set_tau2(float tau2){m_tau2=tau2;}
-  const float tau3() const{return m_tau3;}
+  float tau3() const{return m_tau3;}
   void  set_tau3(float tau3){m_tau3=tau3;}
-  const double chf() const{return m_chf;}
+  double chf() const{return m_chf;}
   void  set_chf(double chf){m_chf=chf;}
-  const double cef() const{return m_cef;}
+  double cef() const{return m_cef;}
   void  set_cef(double cef){m_cef=cef;}
-  const double nhf() const{return m_nhf;}
+  double nhf() const{return m_nhf;}
   void  set_nhf(double nhf){m_nhf=nhf;}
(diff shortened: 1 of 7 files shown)


raggleton:
There are more changes required until code is ready for production, but these are all the HOTVR/XCone-related changes

PR #918, opened 2018-05-17
[80Xv5] Move to ungroomed AK8PUPPI TopJets

raggleton:
As requested, move from `patJetsAk8PuppiJetsSoftDropPacked_daughters` with groomed pt to `packedPatJetsAk8PuppiJets_SoftDropPuppi` with ungroomed pt.

**This is only for AK8 PUPPI, not CHS, etc!** - if requested can also be done for CHS etc.

SoftDrop mass now available via `TopJet::softdropmass()` whilst ungroomed mass available via `TopJet::v4().M()`.
Upside of this is that all the energy fractions, # daughters etc work properly now.
Subjets are there as before.

Also not, I had to add in a "safety factor" to the pt cut on the FastjetJetProducers that ran with SoftDrop to ensure there is always a matching groomed jet for each ungroomed jet.

(Technical note, the use of `JetSubstructurePacker` allows one to store subjet info which can then be accessed using `pat::Jet::subjets("...")`.)

### DIFF
--- a/core/plugins/NtupleWriterJets.cxx
+++ b/core/plugins/NtupleWriterJets.cxx
@@ -324,12 +324,12 @@ NtupleWriterTopJets::NtupleWriterTopJets(Config & cfg, bool set_jets_member): pt
     src_hepTopTagPuppi_token = cfg.cc.consumes<edm::View<reco::HTTTopJetTagInfo> >(edm::InputTag("hepTopTagPuppi"));
 
     pruned_src = cfg.pruned_src;
-    if(pruned_src.find("Mass")==string::npos){
+    if(pruned_src.find("Mass")==string::npos && pruned_src != ""){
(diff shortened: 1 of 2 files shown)


raggleton:
BTW an interesting effect: if you compare ungroomed vs groomed jets on a jet-by-jet basis, the ungroomed jets can actually often have a _lower_ pT than the groomed ones (unexpectedly!). This is *with* JEC applied though - if you compare the _uncorrected_ pT, then ungroomed jets have a larger pT (as exepcted). I guess this is due to the difference in correction factor being larger than the difference made by grooming.

e.g:

| type | uncorrected pt | JEC factor | corrected pt | eta | phi | SD mass |
| --- | --- | --- | --- | --- | --- | --- |
| ungroomed | 179.46603 | 0.9414441 | 168.95724 | -0.607387 | -1.852684 | 49.983230 |
| groomed | 177.84790 | 1.0628090 | 189.01835 | -0.609613 | -1.855665 | 49.983265 |

PR #933, opened 2018-05-30
More generic XConeProducer

denschwarz:
Now you can define the XCone setup (for RECO jets) in the Ntuplewriter.py.
One can set number of jets, cone radius and beta as well as number, radius and beta for subjets. 
If number of subjets are set to 0, no clustering of subjet is performed and empty subjet collections are added to the fatjets. The requirement of sufficient particles is now related to the number of required jets and subjets. 

I also removed a line from the HOTVR in the ntuplewriter where a bool was set which was not used.

I tested this with chs and puppi in 1000 events for both SingleElectron and ttbar.

### DIFF
--- a/core/plugins/XConeProducer.cc
+++ b/core/plugins/XConeProducer.cc
@@ -69,12 +69,12 @@ class XConeProducer : public edm::stream::EDProducer<> {
     edm::EDGetToken src_token_;
     const std::string subjetCollName_;
     const bool usePseudoXCone_;
-    unsigned int NJets_;
-    double RJets_;
-    double BetaJets_;
-    unsigned int NSubJets_;
-    double RSubJets_;
-    double BetaSubJets_;
+    unsigned int NJets_ = 2;
+    double RJets_ = 1.2;
+    double BetaJets_ = 2.0;
+    unsigned int NSubJets_ = 3;
+    double RSubJets_ = 0.4;
+    double BetaSubJets_ = 2.0;
 };
 
 //
--- a/core/plugins/XConeProducer.cc
+++ b/core/plugins/XConeProducer.cc
@@ -225,12 +225,6 @@ XConeProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup)
 
     // check if there are more particles then required subjets
     bool enoughParticles = (particle_in_fatjet.size() > NSubJets_);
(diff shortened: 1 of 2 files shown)


core/plugins/XConeProducer.cc
```diff
@@ -157,122 +169,99 @@ XConeProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup)
...
-  double jet1_area = 0;
-  double jet2_area = 0;
-  //double jet1_area = fatjets[0].area();
-  //double jet2_area = fatjets[1].area();
-
-  // pat-ify jets and subjets
-  auto jetCollection = std::make_unique<pat::JetCollection>();
+    // check if there are more particles then required subjets
+    bool enoughParticles = (particle_in_fatjet.size() > NSubJets_);
+    if (!enoughParticles) {
```
raggleton: I guess it makes more sense to only issue the LogWarning if `doSubjets` is true

core/plugins/XConeProducer.cc
```diff
@@ -157,122 +169,99 @@ XConeProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup)
...
+      edm::LogWarning("InsufficientParticles")
+        << "Not enough particles in fatjet 1 to run second XCone step: "
```
raggleton: Could you change this to `"Not enough particles in fatjet " << i << " to run subjet XCone: "`

core/plugins/XConeProducer.cc
raggleton: Infact, do we even want the warning? I find I get it frequently enough that it is a little annoying, and I don't know if it adds anything useful. As someone who actually knows/uses XCone, if you think its useful then keep it, otherwise let's remove it. Or downgrade it to a LogDebug or something

denschwarz: Yes, I think we won't need the warning, I will remove this LogWarning.

raggleton:
Thanks.

Just tested 300 MC events between before/after these changes, exactly the same distributions

raggleton:
BTW in general I found it was a tad tricky to recompile the code - `make clean` doesn't really clean up things, you have to do a `scram b clean` as well, then do `make`

PR #937, opened 2018-06-08
[94X] Forward-port of HOTVR/XCone changes from 80Xv5

raggleton:
Covers:
- Move to CHS & PUPPI for both XCone/HOTVR
- Allow user to choose PseudoXcone or normal (default = pseudo)
- Allow user to configure XCone plugin settings
- Gen level XCone/HOTVR now doesn't include neutrinos
- Properly handle cases where too few particles in XCone

Also includes equivalent fixes for #935 , #936


raggleton:
Hmm should this go into v1? Or should I start another branch now we've done one processing with v1 branch

PR #939, opened 2018-06-11
 changes for new MET TypeI Correction propagation

multh:


raggleton:
Hi, is this missing some changes? I can see you add `Jet::JER_factor_raw()` but not `set_JER_factor_raw()`. And it doesn't seem to set `JER_factor_raw` anywhere, so I would assume in the `correct_MET` method when it does `auto factor_raw_JER = jet.JER_factor_raw();` it just returns the default value of 0?

multh:
Hi, thanks. I totally missed this point in my first commit.

raggleton:
The `Jet` class is still missing a `set_JER_factor_raw()` setter?

multh:
Hopefully, I didn't miss an other point ;)

raggleton:
Looks ok to me - @karavdin is the logic/physics all correct?

common/src/JetCorrections.cxx
```diff
@@ -279,18 +371,27 @@ std::unique_ptr<FactorizedJetCorrector> build_corrector(const std::vector<std::s
       if(jet.v4().Pt() > 15 && (jet.neutralEmEnergyFraction()+jet.chargedEmEnergyFraction())<0.9){
 	auto factor_raw = jet.JEC_factor_raw();
 	auto L1factor_raw = jet.JEC_L1factor_raw();
+	auto factor_raw_JER = jet.JER_factor_raw();
 
-	corrector_L1RC.setJetPt(jet.pt() * factor_raw);
+	corrector_L1RC.setJetPt(jet.pt() * factor_raw_JER);
```
karavdin: setJetPt should take raw jet pt. This is not raw, but L123 corrected, right?

common/src/JetCorrections.cxx
```diff
@@ -279,18 +463,27 @@ std::unique_ptr<FactorizedJetCorrector> build_corrector(const std::vector<std::s
...
+	corrector_L1RC.setJetE(jet.energy() * factor_raw_JER);
 	corrector_L1RC.setJetA(jet.jetArea());
 	corrector_L1RC.setRho(event.rho);
 	auto correctionfactors_L1RC = corrector_L1RC.getSubCorrections();
 	auto correctionfactor_L1RC  = correctionfactors_L1RC.back();
 
-
-	LorentzVector L1corr =   (L1factor_raw*factor_raw)*jet.v4();            //L1 corrected jets
-	LorentzVector L1RCcorr = (correctionfactor_L1RC*factor_raw)*jet.v4();   //L1RC corrected jets
+	corrector.setJetPt(jet.pt() * factor_raw);
```
karavdin: Also here RAW jet pt should be used as input jet.pt() * factor_raw would be L123 corrected jet pt

common/src/JetCorrections.cxx
```diff
@@ -279,18 +463,27 @@ std::unique_ptr<FactorizedJetCorrector> build_corrector(const std::vector<std::s
       if(jet.v4().Pt() > 15 && (jet.neutralEmEnergyFraction()+jet.chargedEmEnergyFraction())<0.9){
 	auto factor_raw = jet.JEC_factor_raw();
 	auto L1factor_raw = jet.JEC_L1factor_raw();
```
karavdin: Where JEC_factor_raw is set now?

common/src/JetCorrections.cxx
karavdin: and for corrector.setJetPt(jet.pt() * factor_raw); it should be jet.pt() * factor_raw * factor_raw_JER

common/src/JetCorrections.cxx
```diff
@@ -461,31 +461,32 @@ std::unique_ptr<FactorizedJetCorrector> build_corrector(const std::vector<std::s
...
 	auto correctionfactor  = correctionfactors.back();
 
-	LorentzVector L1corr =   (L1factor_raw*factor_raw_JER)*jet.v4();            //L1 corrected jets
-	//	LorentzVector L1RCcorr = (correctionfactor_L1RC*factor_raw_JER)*jet.v4();   //L1RC corrected jets 
-       	LorentzVector L1RCcorr = (correctionfactor_L1RC/correctionfactor)*jet.v4();   //L1RC corrected jets Update 15 May 2018
+	LorentzVector L1corr =   (L1factor_raw * factor_raw_JEC)*jet.v4();            //L1 corrected jets
+       	LorentzVector L1RCcorr = (correctionfactor_L1RC * factor_raw_JEC )*jet.v4();   //L1RC corrected jets Update 15 May 2018
 	LorentzVector L123corr = jet.v4();                                      //L123 corrected jets (L23 in case of puppi)
 
+	cout<<"inv. correction_factor: "<<1/correctionfactor<<",    factor_raw_JEC: "<<factor_raw_JEC<<endl;
```
raggleton: Please remove the cout

multh:
Anastasia, 
could you please check carefully, that I used now the right factors on the right jets? 
The code compiles and I checked, that the correction factor I calculated and the one I already have are the same.

common/src/JetCorrections.cxx
```diff
@@ -268,31 +453,31 @@ std::unique_ptr<FactorizedJetCorrector> build_corrector(const std::vector<std::s
   //propagate to MET
   //apply type1 MET correction to RAW MET
   //NB: jet with substracted muon Pt should be used
-
-  void correct_MET(const Event & event, const bool & do_L1corr, const bool do_L1RCcorr, FactorizedJetCorrector & corrector_L1RC){
+ void correct_MET(const Event & event, const bool & do_L1corr, const bool do_L1RCcorr, FactorizedJetCorrector & corrector_L1RC, FactorizedJetCorrector & corrector){
```
raggleton: Does this mean you don't need the `corrector` arg anymore?

multh: Yes, thats true. Totally missed this point. Thanks.

karavdin:
@multh, code looks good. Could you create PR for 94X branch too, please?

karavdin:
@raggleton, we should propagate these changes to 80X_v5, 94X_v2 and future 10X

PR #962, opened 2018-06-26
[80v5] GenHOTVR/XCone producers

raggleton:
This moves the GenHOTVR and GenXCone clusterers into their own modules, out of UniversalGenCluster and Ntuplewriter. It also allows more flexibility in future with regards to inputs/configuration, and fewer bottlenecks when running multicore.

All the previous functionality is the same except:

- I have removed the softdrop mass from genxcone as (a) it wasn't being used, (b) it requires more work to setup correctly to get the subjets. Let me know if you want it back in, it can be done.
- The GenXConeProducer has the ability to find the highest pt lepton, and reduce the NJets by 1 for the jet closest to the lepton (as before), except now it isn't hardcoded for NJets=3, and you can also set a maximum DeltaR for a lepton to be considered matched to a jet.

One question though for @denschwarz, should the lepton be removed from list of constituents before the subjet clustering is done? At the moment it's left in, like in the UniversalGenClusterer.

I tested on TTbar MC. 

Some timings for 300 events:
For HOTVR:
```
TimeReport   0.162737     0.162737     0.162737  hotvrCHS
TimeReport   0.003650     0.003650     0.003650  hotvrGen
TimeReport   0.077250     0.077250     0.077250  hotvrPuppi
```

For XCone:
```
TimeReport   0.100796     0.100796     0.100796  xconeCHS
TimeReport   0.089956     0.089956     0.089956  xconeGen23Lepton
TimeReport   0.092013     0.092013     0.092013  xconeGen33
TimeReport   0.089104     0.089104     0.089104  xconePuppi
```

The ntuplewriter:
```
TimeReport   0.007634     0.007634     0.007634  MyNtuple 
```

and in total:

```
TimeReport       event loop CPU/event = 2.111039
TimeReport      event loop Real/event = 2.152348
TimeReport     sum Streams Real/event = 2.148585
TimeReport efficiency CPU/Real/thread = 0.980808
```

Weirdly, the Ntuplewriter went from ~0.02s/event to 0.007 pulling out the gen clustering. But the new modules take much longer than ~0.013s, so bit confused where the loss in speed is.

Also it failed to find a lepton in 10/300 events, which I guess would be where you have either a tau decay, or all-hadronic scenario?

Will also port to master, 94, 101

### DIFF
--- a/core/plugins/GenHOTVRProducer.cc
+++ b/core/plugins/GenHOTVRProducer.cc
@@ -1,0 +1,231 @@
+// -*- C++ -*-
+//
+// Package:    UHH2/GenHOTVRProducer
+// Class:      GenHOTVRProducer
+//
+/**\class GenHOTVRProducer GenHOTVRProducer.cc UHH2/GenHOTVRProducer/plugins/GenHOTVRProducer.cc
+
+ Description: [one line class summary]
+
+ Implementation:
+    [Notes on implementation]
+*/
+//
+// Original Author:  Robin Aggleton
+//         Created:  Tue, 23 Jan 2018 13:27:50 GMT
+//
+//
+
+
+// system include files
+#include <memory>
+
+// user include files
+#include "FWCore/Framework/interface/Frameworkfwd.h"
+#include "FWCore/Framework/interface/stream/EDProducer.h"
+
+#include "FWCore/Framework/interface/Event.h"
+#include "FWCore/Framework/interface/MakerMacros.h"
+
+#include "FWCore/ParameterSet/interface/ParameterSet.h"
+#include "FWCore/Utilities/interface/StreamID.h"
+
+#include "DataFormats/Candidate/interface/Candidate.h"
+#include "DataFormats/PatCandidates/interface/Jet.h"
+#include "DataFormats/Math/interface/LorentzVector.h"
+
+#include "fastjet/ClusterSequence.hh"
+#include "fastjet/ClusterSequenceArea.hh"
+#include "fastjet/contrib/HOTVR.hh"
+#include "fastjet/contrib/HOTVRinfo.hh"
+#include "fastjet/contrib/Nsubjettiness.hh"
+#include "fastjet/contrib/XConePlugin.hh"
+#include "fastjet/contrib/SoftDrop.hh"
+
+#include "vector"
+
+using namespace fastjet;
+using namespace contrib;
+
+//
+// class declaration
+//
+
+class GenHOTVRProducer : public edm::stream::EDProducer<> {
+  public:
+    explicit GenHOTVRProducer(const edm::ParameterSet&);
+    ~GenHOTVRProducer();
+
+    static void fillDescriptions(edm::ConfigurationDescriptions& descriptions);
+
+  private:
+    virtual void beginStream(edm::StreamID) override;
+    virtual void produce(edm::Event&, const edm::EventSetup&) override;
+    virtual void endStream() override;
+
+    virtual pat::Jet createPatJet(const PseudoJet &);
+
+    // ----------member data ---------------------------
+    edm::EDGetToken src_token_;
+    const std::string subjetCollName_;
+    double mu,                 // massjump threshold
+      theta,                   // massjump parameter
+      max_r,                   // maximum allowed distance R
+      min_r,                   // minimum allowed distance R
+      rho,                     // cone shrinking parameter
+      hotvr_pt_min;            // minimum pT of subjet
+};
+
+//
+// constructors and destructor
+//
+GenHOTVRProducer::GenHOTVRProducer(const edm::ParameterSet& iConfig):
+  src_token_(consumes<edm::View<reco::Candidate>>(iConfig.getParameter<edm::InputTag>("src"))),
+  subjetCollName_("SubJets"),
+  mu(iConfig.getParameter<double>("mu")),
+  theta(iConfig.getParameter<double>("theta")),
+  max_r(iConfig.getParameter<double>("max_r")),
+  min_r(iConfig.getParameter<double>("min_r")),
+  rho(iConfig.getParameter<double>("rho")),
+  hotvr_pt_min(iConfig.getParameter<double>("hotvr_pt_min"))
+{
+  produces<pat::JetCollection>();
+  produces<pat::JetCollection>(subjetCollName_);
+}
+
+
+GenHOTVRProducer::~GenHOTVRProducer()
+{
+
+  // do anything here that needs to be done at destruction time
+  // (e.g. close files, deallocate resources etc.)
+
+}
+
+
+//
+// member functions
+//
+
+// ------------ method called to produce the data  ------------
+void
+GenHOTVRProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup)
+{
+  // Set the fastjet random seed to a deterministic function
+  // of the run/lumi/event.
+  // NOTE!!! The fastjet random number sequence is a global singleton.
+  // Thus, we have to create an object and get access to the global singleton
+  // in order to change it.
+  // Taken from VirtualJetProducer
+  fastjet::GhostedAreaSpec gas;
+  std::vector<int> seeds(2);
+  unsigned int runNum_uint = static_cast <unsigned int> (iEvent.id().run());
+  unsigned int evNum_uint = static_cast <unsigned int> (iEvent.id().event());
+  uint minSeed_ = 14327;
+  seeds[0] = std::max(runNum_uint, minSeed_ + 3) + 3 * evNum_uint;
+  seeds[1] = std::max(runNum_uint, minSeed_ + 5) + 5 * evNum_uint;
+  gas.set_random_status(seeds);
+
+  edm::Handle<edm::View<reco::Candidate>> particles;
+  iEvent.getByToken(src_token_, particles);
+
+  // Convert particles to PseudoJets
+  std::vector<PseudoJet> _psj;
+  for (const auto & cand: *particles) {
+    if (std::isnan(cand.px()) ||
+        std::isnan(cand.py()) ||
+        std::isnan(cand.pz()) ||
+        std::isinf(cand.px()) ||
+        std::isinf(cand.py()) ||
+        std::isinf(cand.pz()) ||
+        (cand.pt() == 0))
+      continue;
+
+    _psj.push_back(PseudoJet(cand.px(), cand.py(), cand.pz(), cand.energy()));
+  }
+
+  // Do the clustering, make jets, nsub, store
+  // jet definition with HOTVR plugin
+  HOTVR hotvr_plugin(mu, theta, min_r, max_r, rho, hotvr_pt_min, HOTVR::CALIKE);
+  JetDefinition jet_def(&hotvr_plugin);
+
+   // Setup cluster sequence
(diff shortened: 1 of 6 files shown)


raggleton:
@denschwarz I realise I actually added in the area calculation here as well, like in the reco XConeProducer. Is this needed? If not, I can remove it and save a lot of time (we're talking about going from ~0.09s /evt to ~0.005s/evt for each GenXCone module)

denschwarz:
Concerning the lepton:
For my special case the lepton is actually part of the jet, so it would be nice to have this in the GenJet as well. Are all leptons usually removed in GenJets?

About the jet area:
We are using the jet area only to apply proper L1 corrections. So, this is not needed for GenJets.

raggleton:
After discussion with Dennis, some outstanding changes to be made:

- remove area calculation (gen only)
- update check on min # particles
- leave the lepton in

Then I'll make a small sample for him for both old & new setups to run over and check it looks sane

raggleton:
One related comment about XCone area: for reco XCone area, I notice that the fatjet area is set to 0, but we do store the subjet area. Should the fatjet area also be stored?

denschwarz:
Again, for my purpose this is perfectly fine because I am using and correcting the subjets only. So, for timing reasons I would only store the subjet area in the current setup.

raggleton:
Ntupleised ~25K TTbar events with this change for you to try. Files here: /pnfs/desy.de/cms/tier2/store/user/raggleto/TT_TuneCUETP8M2T4_13TeV-powheg-pythia8/crab_TT_TuneCUETP8M2T4_13TeV-powheg-pythia8testXCone/180627_091918/0000/

raggleton:
Note to future self: although I measure the average event running interactively on NAF as ~ 1.9s /evt, on CRAB I never get this fast. From 11 jobs I get:

```
cmsRun-stdout-1.log:TimeReport      event loop Real/event = 2.734822
cmsRun-stdout-10.log:TimeReport      event loop Real/event = 2.854104
cmsRun-stdout-11.log:TimeReport      event loop Real/event = 2.878038
cmsRun-stdout-2.log:TimeReport      event loop Real/event = 2.448509
cmsRun-stdout-3.log:TimeReport      event loop Real/event = 1.965823
cmsRun-stdout-4.log:TimeReport      event loop Real/event = 2.922009
cmsRun-stdout-5.log:TimeReport      event loop Real/event = 2.592143
cmsRun-stdout-6.log:TimeReport      event loop Real/event = 2.866816
cmsRun-stdout-7.log:TimeReport      event loop Real/event = 3.063018
cmsRun-stdout-8.log:TimeReport      event loop Real/event = 2.278690
cmsRun-stdout-9.log:TimeReport      event loop Real/event = 2.410607
```

denschwarz:
Perfekt, I will run my analysis over your sample of ttbar events!

denschwarz:
I just wanted to run your set of ttbar Ntuples but there is a naming issue.
So, the names of the GenJet Collections changed. I think this is not a good idea regarding the 80X_v5 version. It would be quite unpractical to deal with different collection names in different samples. And this also means that I cannot run my analysis out of the box.

If you would like to keep the names (and only implement the changes in newer versions) I could set up a working analysis for testing but maybe it would be faster to change the names and produce a small set of events again? Then, a test with the current version of my analysis code would also be a consistency check with the samples we already produced.

The names that changed are:
'genXCone33TopJets' changed to 'xconeGen33'
'genXCone23TopJets' changed to 'xconeGen23Lepton'
'genXCone33TopJets_softdrop' is not included anymore but this is fine.

raggleton:
OK, made new set of test files: /pnfs/desy.de/cms/tier2/store/user/raggleto/TT_TuneCUETP8M2T4_13TeV-powheg-pythia8/crab_TT_TuneCUETP8M2T4_13TeV-powheg-pythia8testXCone2/180627_133524/0000/

denschwarz:
Alright, I ran some numbers now. You created 25230 events, XCone finds 50460 fatjets in both categories which is expected if no event has a very small number of GenParticles.

Now going to subjets. In the 33 method we should find a maximum of 50460*3 = 151380 jets. We are actually storing 151057 jets, which means that ~0.2% of the fatjets do not have subjets (which is kind of reasonable). What puzzles me is that  151057 is not dividable by 3. So it is not either 3 or 0 subjets in this method.

In the 23 method we should find a max of 50460*(5/2) = 126150 jets. We are actually storing 0.6% jets MORE. I think this can be explained because you changed the way the clustering behaves if no lepton is found. In this case, we are clustering 3 subjets in each jet.

So, I am really wondering about this 33 method that there are not N*3 subjets. Could it happen, that we have enough particles in the fatjet, but one (or more) subjet ends up with no particles inside and is somehow not stored?

raggleton:
Just ran a quick test, and seen one event where we ask for 3 subjets but it only gives 2 subjets (Event: 794651 in `/store/mc/RunIISummer16MiniAODv2/TT_TuneCUETP8M2T4_13TeV-powheg-pythia8/MINIAODSIM/PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6-v1/80000/6ADE8687-9DBE-E611-AB0F-001E677924C6.root`)

```
Only found 2 subjets but requested 3. Fatjet had 16 constituents
pt : eta : phi
fatjet: 81.8495 : -1.19804 : 3.58054
subjet: 78.4041 : -1.16899 : 3.58434
subjet: 1.37493 : -1.35413 : 2.47644
```

denschwarz:
I had a look into RecoJets now. And here we see exactly 3*N_fat subjets. This is true for your Ntuples and also for Ntuples from the last production (no PseudoXCone and still using UniversalJetCluster).

raggleton:
I just ran with the old UniversalGenCluster over 2 events where you get 2 subjets instead of the 3 requested (`1:794651`, `1:794709`), and see the exact same fatjets & subjets in both modules. So it appears this issue did exist before.

As an extra check, I ran the old UniversalGenCluster with the packedGenParticles (so including neutrinos), and I still get only 2 subjets (albeit with different jet/subjet pt/eta/phi).

raggleton:
OK so for an example event:

```
fatjet: 81.8495 : -1.19804 : 3.58054
fatjet constitutent: 0.748535 : -1.20058 : 4.73919
fatjet constitutent: 0.353516 : -1.87448 : 3.74551
fatjet constitutent: 0.588867 : -0.501423 : 3.72119
fatjet constitutent: 0.568359 : -1.23126 : 2.43659
fatjet constitutent: 0.494385 : -1.80673 : 2.91445
fatjet constitutent: 0.448242 : -1.8631 : 3.98644
fatjet constitutent: 45.9687 : -1.16715 : 3.59023
fatjet constitutent: 0.407715 : -1.56597 : 2.58211
fatjet constitutent: 0.4021 : -1.27262 : 2.42576
fatjet constitutent: 0.293213 : -0.950093 : 4.36926
fatjet constitutent: 0.217651 : -0.950658 : 3.08447
fatjet constitutent: 0.310303 : -1.52337 : 4.14796
fatjet constitutent: 0.102722 : -0.518387 : 3.4622
fatjet constitutent: 17.7656 : -1.17045 : 3.57236
fatjet constitutent: 14.6719 : -1.17283 : 3.58037
fatjet constitutent: 0.0427856 : -0.803858 : 2.90722
subjet: 78.4041 : -1.16899 : 3.58434
subjet constitutent: 45.9687 : -1.16715 : 3.59023
subjet constitutent: 17.7656 : -1.17045 : 3.57236
subjet constitutent: 14.6719 : -1.17283 : 3.58037
subjet: 1.37493 : -1.35413 : 2.47644
subjet constitutent: 0.568359 : -1.23126 : 2.43659
subjet constitutent: 0.407715 : -1.56597 : 2.58211
subjet constitutent: 0.4021 : -1.27262 : 2.42576
```

So for sure, it's not using all the constituents...very odd

denschwarz:
Could you check the jets clustered with the 23 method in the same event?

raggleton:
BTW this is what that event looks like: blue is the fatjet constituents, red is the high pt subjet, green the other. The diamonds are the jet axes. The rings are radii of 1.2 for fatjet, 0.2 for subjets. Fatjet constituents are sized according to their pT. So you can see that basically all the pT is in those few particles at the middle.


image

raggleton:
And if you look at the 23 method, you get exactly the same jet (when asking for 2 subjets), with exactly the same subjets and constituents

denschwarz:
Just a small command regarding your plot:
Subjet radii should be 0.4, but of course the statement does not change.

From the numbers and the plot I would guess that these two jets are indeed the most energetic you can get in this event. But I cannot see why a very low-energetic jet is not stored (even a third jet just containing the most energetic but not clustered particle would have 0.7 GeV). Maybe there is a 'magic' 1GeV boundary somehow?

denschwarz:
Did you find some other example events we could use to find a pattern?

raggleton:
Reminder to self on status: trying to investigate why sometimes we don't get exactly N jets, even though docs say that we should - should provide 4-vecs with 0 pT if unable to find more jets. Seems to happen with full plugin as well as pseudo plugin. And why it won't find more subjets even though there are more particles there (see diagram above - only 2 subjets when requested 3, even though other particles about. Some hidden pT cut?)

raggleton:
I think the modules are basically ready - the one issue with GenXCone producing < N jets is an issue with XCone plugin (I think). I'm not sure whether to add in a warning/error in those situations to alert the user.

There was one other consideration for these Gen modules - at the moment, they do not calculate jet area. Calculating jet area is where all the slowness happens, and is currently not needed for XCone/HOTVR studies as I understand. However, you mentioned an interest in using GenXCone - would you need the genjet area? I was thinking of perhaps adding a bool to the module, and then calculate the area only if the user requests it. But if no one ever needs it, then I won't bother with it.

karavdin:
I guess jet area is needed only for calculation of offset jet energy 
correction. For gen jets we won't need it, so  it's ok to have GenXCone 
without jet area stored.
For the issue with GenXCone producing < N jets, I would add warning so 
user is aware of such cases.

raggleton:
In terms of the underlying bug in XCone/Njettiness: I've been digging around, and have found one bug/feature.

In `NjettinessPlugin::run_clustering` in `Nsubjettiness/NjettinessPlugin.cc`,  it loops through all the axes returns by the Nsubjettiness procedure, and how many particles are associated with each axis. If there are 0, then it doesn't add that axis to the output, hence we see fewer than N subjets.

*But* the next confusing thing is that before this step, in an example "bad" Event, the axis _does_ have particles associated with it, which then fail some distance test, which I don't understand. To be continued...

raggleton:
For our producers, I have now added in both a warning to the user that they haven't got the number of subjets requested, as well as adding in blank jets to make it up to the number requested. @denschwarz is this "good" behaviour? I thought it might be better this way, since one might assume that calling XCone with N guarantees N jets (even if some are pt=0).

denschwarz:
So you are adding 4-vectors with all entries =0 for every 'missing' jet? This seems like a good behaviour because it is unlikely to see all the warnings and this way you can scroll through and clean your jet collection on analysis level.

raggleton:
Right, think this is finally finished. I've added in a number-of-jets check for both fatjets and the subjets (if the user wanted subjets), for both the Reco & Gen XCone plugins. If there aren't enough jets, it pads it out with null 4-vectors to make it up to N jets.

denschwarz:
Yes, this sound great! I think everything else should be done on analysis level.

PR #965, opened 2018-07-10
Store additional info for prefiring studies

karavdin:
add L1 seeds and prescales (HLT and L1min, L1max)


core/plugins/NtupleWriter.cc
```diff
@@ -1315,21 +1341,33 @@ bool NtupleWriter::filter(edm::Event& iEvent, const edm::EventSetup& iSetup) {
...
+    auto & triggerPrescalesL1min = *event->get_triggerPrescalesL1min();
+    auto & triggerPrescalesL1max = *event->get_triggerPrescalesL1max();
+
     //muGT 
     edm::Handle<BXVector<GlobalAlgBlk>> l1GtHandle;
     iEvent.getByToken(l1GtToken_, l1GtHandle);
     bool prefire = l1GtHandle->begin(-1)->getFinalOR();
     triggerNames_outbranch.push_back("muGT_BX_minus_1__prefire");
     triggerResults.push_back(prefire);
+    triggerPrescales.push_back(1);
```
raggleton: Is there a reason for storing 1s for all these prescales?

karavdin: Yes, in the code it assumes that triggerResults and triggerPrescales size is the same. Dummy values are stored to avoid changes in the code  and keep it compatible with old ntuples.

PR #967, opened 2018-07-12
Run ii 94 x v1

garbersc:
added V11 and V12 JEC paths


karavdin:
Fall17_17Nov2017_V12_MC does not exist. Please remove it from the JetCorrector class.

PR #977, opened 2018-07-20
[10_1] Update config with GT/files

raggleton:
Also now include ElectronIdentification in install.sh - not sure if actually needed, but had issue with MVA without it?


raggleton:
With this, the ntuplewriter now produces ntuples. (Some collections will need updating, but doesn't look unreasonable at first glance)

PR #978, opened 2018-07-27
bug fixed

anmalara:
MCPileupReweight would crash if the trueNumInteractions is negative: (look at 147-157 lines)


raggleton:
OK, guess this needs porting to master & 10_1 as well.

BTW in future, if you want to make whitespace changes could you do it in a separate PR? This PR suffered as the actual logic changes is a few lines hidden in a lot of other whitespace changes.
If you only want to add part of your changes to a commit, please use `git add -p` and then you can pick which bits you want.

PR #992, opened 2018-08-16
bug fixed in trigger pass

anmalara:
This commit is to raise the same error: assertion error ->runtime error. It's easier to catch.
In the past runtime error was raised if the trigger Name was not present since the first event of the sample. Whereas assertion occurs if the trigger name was not present in the middle of the sample. 
Now both cases raise a runtime error.


raggleton:
OK. Will need porting to master, 10_1 branches

PR #996, opened 2018-08-17
Add MC_QCD_*_EMEnriched samples

MatthiesC:
Added xml files for QCD EMenrichted samples


MatthiesC:
Also added xml files for MC QCD bottom+charm-->electron

PR #1005, opened 2018-09-05
[101] Function-ise ntuplewriter

raggleton:
Proposal to replace single `ntuplerwriter.py` config with a python module that has a function that generates a `cms.Process` object with all the ntuple modules setup (i.e. the contents of ntuplewriter.py), but with args for `useData`, etc. Then we have separate config files for data, MC, with extra HOTVR/XCONE modules, whatever.  

Advantages:

- With more variations of configs used (without XCONE/HTOVR, with extra modules, etc), it is far easier to have the main ntuple contents centrally determined, then customised afterwards in specific concrete configs. These specific configs are then not thousands of lines long, but succinctly shows the differences
- Can easily generate configs for data, MC, for testing purposes in the future - at the moment once you import the ntuplewriter you're stuck with whatever the data switch is set to
- Slightly easier to switch for data & MC CRAB jobs because they would use different config files, rather than remembering to flip the `useData` switch

Disadvantages:

- Extra layer of abstraction/complication?
- Got to ensure that the specific configs don't end up with too much in them that should instead be centralised in the generating function
- **Does this screw up usage with grid-control?** I see there are gc-specific bits in the ntuplewriter, but I have no clue how they're used

An alternate approach could be to use `varParsing` and pass in configurable args on the commandline (and use `pyCfgArgs` in CRAB jobs). Downside of that approach is that you can't directly import the `process` object into other python files

If this seems OK, I will add it into master as well, but not into 94X or older (since they are essentially ~final) unless they are particularly valuable.


karavdin:
Hi Robin,
Many thanks for working on this, should make our live easier. The code looks fine to me.
Could you point to gc-specific bits in the old ntuplewriter?

raggleton:
Speaking with Christoph, it seems that the GC-specific part (https://github.com/UHH2/UHH2/blob/master/core/python/ntuplewriter.py#L78..L90) is not used at the moment, so we'll ignore it for now (and fix it when someone needs it).

PR #1010, opened 2018-09-19
[master] Update install script for HOTVR + other niceties

raggleton:
The fastjet-contrib SVN is no longer reachable by anonymous users (i.e us). I have copied the HOTVR contents to a github repo: https://github.com/UHH2/HOTVRContrib, and this updates the install script to now use that repo as the source of HOTVR.

(cf https://discourse.phabricator-community.org/t/public-svn-fails/1917)

Related, I have also turned on the option that if a command fails, the whole install script fails. e.g. the failure to checkout from svn was not visible, only when do `scram build` when it couldn't find `HOTVR.hh`

Unrelated, I have also taken the opportunity to disable the use of auto-ptr in fastjet. As far as I can tell, using auto-ptr was a backwards-compatible option that we don't actually need. This removes all the warnings that come when compiling.
Also added in the python extension, required in 10_1 and future.

There are also separate PRs for 101, 94_2. 80X uses a central installation of fastjet from my AFS, so isn't affected.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -6,6 +6,8 @@
 #
 # before running this script
 
+# Exit script if command fails
+set -e
 
 # Create default make args for parallel jobs
 if [ -z "$MAKEFLAGS" ]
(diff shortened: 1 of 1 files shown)


raggleton:
Also updated versions of software here to match that in 10_1 branch for consistency

PR #1012, opened 2018-09-19
[94_v2] Update install script for HOTVR + other niceties

raggleton:
Port of #1010 for 94_v2


raggleton:
Can't disable auto-ptr for 9_4_X as RecoJets/JetAlgorithms/src/Qjets.cc uses a `auto_ptr`-specific version of a method (which was fixed for 10_1, so that doesn't need changing)

PR #1016, opened 2018-09-26
Run ii 94 x v2

garbersc:
fixed a bug in the preprocessing macro of JetCorrections.cxx and added V24,V25,V26 JEC definitions


raggleton:
Compilation test running at https://gitlab.cern.ch/raggleto/UHH2-integration/pipelines/522803

raggleton:
Compilation succeeded.

But I did get the warning (that already exists before this PR):
```
src/JetCorrections.cxx: In function 'void __static_initialization_and_destruction_0(int, int)':
src/JetCorrections.cxx:1445:1: note: variable tracking size limit exceeded with -fvar-tracking-assignments, retrying without
 }
 ^
```
that I've not seen before. Not sure if related to worker node it's running on? Or the long #define? Shouldn't affect running.

PR #1017, opened 2018-09-26
Fixed bug in PU reweighting, updated Muon SF files

reimersa:
- Modules would throw when running PU reweighting on data -- instead of just doing nothing.
- Updated muon ID and isolation SF files, they now contain proper systematic uncertainties. Adapted the corresponding function in MCWeight.cxx to new format. Small change in everyone's code needed. 
- Please note that no flat uncertainty should be applied for muon ID and isolation scale factors in 2017 anymore.


reimersa:
- The Fall17 cut-based electron ID in UHH2 had a serious bug (required C0 > XXX instead of C0< XXX), causing most electrons to be cleaned from the collection. 
- EGM apparently released a newly tuned ID here: https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedElectronIdentificationRun2#Working_points_for_94X_and_later that has to be used for analyses after Moriond18
- The pre-computed electronID tags in our ntuples are **outdated**. There is a recipe on the EGM twiki how to obtain the updated tags for 94X via git cms-merge-topic.

karavdin:
Hi Arne,
Thanks for the fix. Could you update electronID tags in 94X ntuple-writer? The next production round is going to happen rather soon.
Cheers,
Anastasia

PR #1033, opened 2018-10-22
Run ii 94 x v2/3

garbersc:
added versions V23-V28 to JetCorrections.

Is there a difference between the *_v2 and *_3 branch already?
Am I right to push and switch to th *_v3 branch?


karavdin:
Please do PR to RunII_94X_v2

PR #1036, opened 2018-10-25
added V32 to JetCorrections

garbersc:
added V32 to JetCorrections


karavdin:
it's 31, 32 does not exist yet. Anyway, the code is correct, merging :)

PR #1037, opened 2018-10-25
V31 in JetCorrections now with DE combined

garbersc:
I forgot, that I had to change the filenames for DE combined...

### DIFF
--- a/common/include/JetCorrections.h
+++ b/common/include/JetCorrections.h
@@ -89,6 +89,23 @@ namespace JERFiles{
   extern const std::vector<std::string> tag##_V##ver##_L123_##jetCollection##_MC;\
   extern const std::vector<std::string> tag##_V##ver##_L1RC_##jetCollection##_MC;\
 \
+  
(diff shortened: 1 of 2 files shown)


garbersc:
removed 2016 JEC definitions and split Data and MC. @karavdin Is that what you meant yesterday?

garbersc:
sorry for merging before letting you check the code. I hope everything is fine

karavdin:
@garbersc , I belive you don't need MC in DEFINE_JERFILES_DEcombined:
https://github.com/UHH2/UHH2/blob/RunII_94X_v2/common/include/JetCorrections.h#L60-L61

Could you also re-name "DEFINE_JERFILES_DEcombined" to "DEFINE_JERFILES_DATA_DEcombined", please?

anmalara:
Another comment. Since those are temporary tests, it's more straightforward to leave the same definition of all the run, and apply the same correction to the multiple runs. ie, runCD with correction for CDcombined, should go to runC,runD separately both applying the correction for CDcombined. you can also include runCD. this prevents changes in the module we are running. Thanks

PR #1045, opened 2018-11-14
[102X] Setup install script, fixes for C++17

raggleton:
Move to CMSSW_10_2_6, latest version. Also move to C++17 & GCC7, and fixes to make code compile under GCC7.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -107,7 +107,7 @@ endif
 
 
 LIBS := $(shell root-config --libs) -lSFrameCore -lXMLParser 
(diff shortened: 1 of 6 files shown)


raggleton:
Note that you also need the latest version of SFrame (updated today), which also has fixes to work in GCC7.

raggleton:
Note that this doesn't have any of the potentially required merge-topics for IDs etc yet (although I'm hoping they're in the CMSSW release already... 🤞)

PR #1048, opened 2018-11-15
[102X] Remove unnecessary cms-addpkg

raggleton:
These were originally needed when CMSSW didn't support
fastjet >=3.2.1. Now it does, so we don't need these packages.


raggleton:
I also ran `nutplewriter_mc.py` and it ran fine with no library errors, so am happy with this

PR #1056, opened 2018-11-23
Added xml for WZ, HZ and ZZ for MC 2017

anmalara:


raggleton:
Cancelled tests as only adding XML files (and the testing isn't smart enough yet to detect that)

PR #1058, opened 2018-11-26
JER SF V3 derived from Pythia and Madgraph samples

anmalara:


raggleton:
We can ignore the failure, it's due to some EE2017 stuff in a config that shouldn't be in this branch. Your actual changes compiled fine 👍

PR #1070, opened 2018-12-11
[102X] CRAB tarball creator & checker, updates to DasQuery

raggleton:
Add small script to crate tarball as CRAB would, and optionally check the size.

Allows us to check we have sensible tarball sizes for CRAB jobs before submitting a load of CRAB jobs.


raggleton:
Trying to test a new test, so please ignore all the failures

PR #1071, opened 2018-12-11
[102X] Update example files in 2016v2 data cfg

raggleton:
Use 03Feb2017 files for data 2016v2


raggleton:
FYI, this was over 100 events with 0 differences in the configs run... so those differences in timings are solely down to low statistics & machine variations.

PR #1072, opened 2018-12-12
[102X] Comment out wrong L1 prefire file location

raggleton:
This was erroneously uncommented during CRAB testing


raggleton:
Actually passed, it says failed because the the reference jobs (ie. current RunII_102X_v1) fails, which is what we're trying to fix...

PR #1073, opened 2018-12-13
[102X] Cleanup trigger obj in ntuple

raggleton:
Remove duplicate triggers

also more frequent printout for logging

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -84,7 +84,7 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=150.):
     task = cms.Task()
 
     process.load("FWCore.MessageService.MessageLogger_cfi")
-    process.MessageLogger.cerr.FwkReport.reportEvery = cms.untracked.int32(1000)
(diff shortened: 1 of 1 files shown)


raggleton:
This isn't a critical update, so I'm using it for some CI testing at the moment, please don't merge

raggleton:
Note to self: 
- timings in https://github.com/UHH2/UHH2/pull/1073#issuecomment-446970010 are for 1K events, and still show (large!) variation even though ~0 difference in configs. Shows the magnitude of variance between machines!

- timings & sizes in https://github.com/UHH2/UHH2/pull/1073#issuecomment-447354937 are for 100 events. Timings are meaningless, but size is decent, and correctly shows a slight reduction since we are removing trigger objects (affects data only).

raggleton:
This PR leads to some slightly odd test results. It is removing duplicate trigger objects, which leads to duplicate collection names. This isn't handled in the plotting scripts, since we assume all collection names are unique to do added/removed collections. The effects of this PR only show up in the file sizes. I am not quite sure how I would detect such duplicate collections to be honest, and its probably more work than reward.

PR #1074, opened 2018-12-17
merge of JEC and JER between branch 94X_v2 and 94X_v3

anmalara:
lots of indentations applied. Sorry about that!

I'd like also to cross-check all the version for JER and JEC present because most of them are not used anymore. Could be useful to think about how to simplify even more the preprocessing directives (maybe for the next iteration).

Once that this is approved, I'd like also to propagate the same changes (SF and new MET) also in 94X_v1 and 94X_v2 branches.

### DIFF
--- a/common/include/JetCorrections.h
+++ b/common/include/JetCorrections.h
@@ -22,7 +22,7 @@ namespace JERFiles{
   extern const std::vector<std::string> tag##_V##ver##_L1RC_##jetCollection##_MC;\
   extern const std::vector<std::string> tag##_V##ver##_L1FastJet_##jetCollection##_MC; \
 \
-  
+
 #define DEFINE_JERFILES_DATA(tag,ver,jetCollection)\
   extern const std::vector<std::string> tag##_V##ver##_B_L123_noRes_##jetCollection##_DATA;\
   extern const std::vector<std::string> tag##_V##ver##_C_L123_noRes_##jetCollection##_DATA;\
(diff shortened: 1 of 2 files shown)


raggleton:
The whitespace probably needed updating at some point anyway... so longs as it's consistently all spaces or all tabs, that's fine.

FYI, this test will say "failed" because it's setup to make ntuples on 10_2_X, but the compilations steps are still valid, will check when done.

raggleton:
re: simplifying the sets used, I would say: keep the newest ones, and check with @garbersc  and @karavdin about which ones are still of interest in each branch for e.g. specific investigations. But a cleanup is certainly welcome 😄 

re: preprocessing directives, it would be interesting to know people's experience with it: does it make life easier? Or too many "special cases" to add in? In 10_2_X, I tried a different approach of moving all the JEC sets to a separate header file, but keeping it "old style" without the preprocessing directives. Not sure which is better, maybe a combo of both? I quite like the idea of separating the logic & implementation vs the names of sets & txt files.
see: https://github.com/UHH2/UHH2/blob/RunII_102X_v1/common/include/JetCorrectionSets.h, https://github.com/UHH2/UHH2/blob/RunII_102X_v1/common/src/JetCorrectionSets.cxx

raggleton:
Compilation actually OK, but a later step in testing failed (unrelated to this PR), so safe to merge

PR #1076, opened 2018-12-18
[94Xv3] CRAB tarball checker + changes for testing

raggleton:
As #1070: crab tarball checker, handles missing datasets better, and if cannot get username from DB, still use default.


raggleton:
OK, think I've figured out where these come from... In 94Xv1/2, to get the EGamma IDs, we do:

PR #1077, opened 2018-12-18
[94Xv2] CRAB tarball checker + changes for testing

raggleton:
As #1076


raggleton:
Restarting tests to get crab tarball listing

PR #1078, opened 2018-12-18
[94Xv1] CRAB tarball checker + changes for testing

raggleton:
As for #1070


raggleton:
OK, this one failed because of a testing step that's 102X-specific. Otherwise OK, although crab tarball is 90MB.

PR #1079, opened 2018-12-18
Run ii 94 x v3

anmalara:
After the discussion of yesterday, I tried to simplify even more the creation of JEC Files.
Both @raggleton and @karavdin, could you please check and tell me if you find it more readable and more accessible to extend?
@karavdin could you cross check I put every relevant version and each correction correctly?


raggleton:
Don't worry about the CRAB tarball check, that's being fixed elsewhere.

Couple of comments:

- You have both JetCorrection**s**Sets.h and JetCorrectionSets.h, but I only see `#include "UHH2/common/include/JetCorrectionsSets.h"`
- Although it compiled, should probably give it a test in a sframe analysis module since compilation doesn't check if the filenames resolve properly...
- ... and I guess that the Summer16 MC won't work (or Fall17 MC), since there is an ultra-subtle underscore difference in filenames (`#tag_V` vs `#tagV` iirc), hence why there where 2 separate (seemingly identical) functions before. Not sure how you'd handle it in this case... maybe an extra "_" after the tag in the argument? But then that also gets propagated to the variable name, which may not desirable...
- Would you mind swapping the massive ASCII text for smaller comments? I find them hard to read. Even better would be an instruction on how a user can add a new set of JECs
- You should sync with the changes made in #1075 in terms of sets we want

Apart from those, it looks quite promising!

anmalara:
> * You have both JetCorrection**s**Sets.h and JetCorrectionSets.h, but I only see `#include "UHH2/common/include/JetCorrectionsSets.h"`

Done!

> * Although it compiled, I should probably give it a test in a sframe analysis module since compilation doesn't check if the filenames resolve properly...

Done! I check on my analysis. It worked for both DATA and MC (note I slight change MC files creation).
Of course one has to be careful in his own code, as for example in the JEC repository V24 has no run DE whereas V31-32 have no D and E separately. But I'd say it's even safer as one has to check the latest recommendations. 

> * ... and I guess that the Summer16 MC won't work (or Fall17 MC), since there is an ultra-subtle underscore difference in filenames (`#tag_V` vs `#tagV` iirc), hence why there where 2 separate (seemingly identical) functions before. Not sure how you'd handle it in this case... maybe an extra "_" after the tag in the argument? But then that also gets propagated to the variable name, which may not desirable...

Done! I prefer to have it separately rather than add "_" somewhere. Thanks about that.


> * Would you mind swapping the massive ASCII text for smaller comments? I find them hard to read. Even better would be an instruction on how a user can add a new set of JECs

Done! Both

> * You should sync with the changes made in #1075 in terms of sets we want

Done!

karavdin:
Hi @anmalara ,
Crushed code due to different DE and D-E split in 32 and earlier version is not safer option for user. This difference was coded internally by experts and still should be done at this level. Please bring back stable behaviour.

anmalara:
> Hi @anmalara ,
> Crushed code due to different DE and D-E split in 32 and earlier version is not safer option for user. This difference was coded internally by experts and still should be done at this level. Please bring back stable behaviour.

I don't think I get your point. Even before one could have had the same problem, if the user programmed your code for different runs, separately.
I want to make clear that there are no compilation errors also in user code, but only runtime errors, given that such files cannot be found. Therefore when the user decides to apply some version JEC should know in advance which runs it has to correct and how (before V31 each separately and after combined).
The first version of the code didn't create at all run D and E separately, but rather DE only. And we agreed to make also files for run D and E separately, each containing DE corrections. Is this what you want again? If so, what's the point of having DE combined and D and E too?

karavdin:
In the currently used set-up (with JetCorrections class only), it's still possible to call RunD and RunE JEC separetly, it will just read them from the same file:
https://github.com/UHH2/UHH2/blob/RunII_94X_v3/common/src/JetCorrections.cxx#L139-L144
No changes needed on the user side and running with V6 and V32 is identical in the AnalysisModule, by calling JEC run by run. The picking up the right txt file is done internally, meaning it's coded in one common place (JetCorrections class at the moment). This is behaviour from common class JetCorrectionsSets I wish to have as a user.

anmalara:
> In the currently used set-up (with JetCorrections class only), it's still possible to call RunD and RunE JEC separetly, it will just read them from the same file:
> https://github.com/UHH2/UHH2/blob/RunII_94X_v3/common/src/JetCorrections.cxx#L139-L144
> No changes needed on the user side and running with V6 and V32 is identical in the AnalysisModule, by calling JEC run by run. The picking up the right txt file is done internally, meaning it's coded in one common place (JetCorrections class at the moment). This is behaviour from common class JetCorrectionsSets I wish to have as a user.

Totally agreed! I still don't get if you prefer to have also DE combined files or not. Given what we said, I assume and support not to have DE combined, but I'd prefer to have confirmation (maybe I don't see the usage) before doing a commit again given that this will change the approach significantly. Thanks :)

reimersa:
I would prefer the name of the JECFile object reflect its actual content instead of just being backward compatible. We could modify CommonModules for that purpose, such that it can still be applied and do the right thing out of the box. 

If users are applying JEC on their own (i.e. without CommonModules), wouldn't it also make sense to give them meaningful JECFile names? Even if that might require a bit of search-replace?

karavdin:
@anmalara, we can skip dedicated combined DE JEC to reduce number of created objects, but one still would need 2 sets of "SET_JERFILES" to handle 2 different cases.
@reimersa, usage of CommonModules is not forced and only JetCorrector is used in any case.

reimersa:
@karavdin, correct! 

So why wouldn't one either use CommonModules for going the easy and less flexible way or apply JEC on their own, using meaningful JECFile names that actually reflect their content as input for the JetCorrectors?

karavdin:
@reimersa, to minimise number of changes. The approach I'm proposing to make the main changes in one place, where it's relevant. The approach with picking up right files by user would require change in at least 2 places: main class to add it and user class  to pick up the right one instead of just changing the name.

PR #1080, opened 2018-12-18
[94Xv2] Remove .git from EGamma external files as inflates CRAB tarball

raggleton:
This is the largest file in the CRAB tarball, making it perilously close to the upload limit of 100MB:
```
33028538 external/slc6_amd64_gcc630/data/RecoEgamma/ElectronIdentification/data/.git/objects/pack/pack-29a80f55679276cfbf198603acd6889a96414597.pack
```
(size in Bytes)

It has no purpose for analysis. And since they now move to v2 IDs (which is in 94X_v3 branch as it requires 9_4_9+), we shouldn't ever need to pull

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -160,9 +160,8 @@ rm TMVA_*.weights.xml
 rm -r Spring16*
 rm -r  Spring15/
 rm -r PHYS14/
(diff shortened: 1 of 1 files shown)


raggleton:
Tarball down to 59MB from 91MB

PR #1081, opened 2018-12-18
[94Xv1] Remove .git from EGamma external files as inflates CRAB tarball

raggleton:
As #1080


raggleton:
Yay tarball down to 59MB (was 91MB)

PR #1082, opened 2018-12-18
[94Xv3] Remove old EGamma ID files during installation to slim CRAB tarball

raggleton:
This removes the following EGamma ID data files since they contribute unnecessarily to the CRAB tarball and make it too large:

- PHYS14
- Spring15
- Spring16
- Summer16

as seen in #1076 .
Note that the only Electron IDs in this branch used are the Fall17 V2 ones, so should be safe.

CRAB tarball drops dramatically in size, I measured it now as 75MB, down from 108MB.


raggleton:
As expected, tarball size is now down to 72MB, well inside the 100MB limit.

PR #1084, opened 2019-01-08
[102X] Better Ntuple plotter & comparison script

raggleton:
This is the new script I use in the automated testing, replacing the existing `compareTrees.py`. It is able to properly loop over all the getters of all classes for all branches, recursively, unlike the previous script. (e.g we can look at the btag info of the subjets of the top jets). This is because previously we only used `TTree::Draw`, which has this inherent limitation. The price for this added complexity is lines of code (by almost an order of magnitude!)
This previous limitation contributed to missing things like subjet variables not being set properly. 

The script can then compare one Ntuple to another, and produce a status report as a JSON file. It can also produce a JSON file of the class methods in one/both files.

Might be useful, and the technique of how to properly iterate through our classes recursively is not so trivial so good to have an example.

Doesn't need CI testing as it is just copying over the script from the CI tests


raggleton:
Failure was in a CI script that wrote a github comment. Will manually override and merge

PR #1085, opened 2019-01-11
[102X] Add 2016 electron IDs back in

raggleton:
Adds back in the electron IDs from 80X branch:

- heepElectronID_HEEPV60
- cutBasedElectronID_Summer16_80X_V1_veto
- cutBasedElectronID_Summer16_80X_V1_loose
- cutBasedElectronID_Summer16_80X_V1_medium
- cutBasedElectronID_Summer16_80X_V1_tight
- cutBasedElectronHLTPreselection_Summer16_V1

(see https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedElectronIdentificationRun2#Working_points_for_2016_data_for)

+ the MVA values for GeneralPurpose and HZZ, and the correct eff area file.

I also updated the ExampleElectronID module with the new IDs, and also added in plots for the MVA values to check they're OK.
There's also some whitespace tidying up.

Some notes/questions:
- HEEPv7 is the one now recommended for 2016 and 2017/8, but it seems we had HEEPv6 in the 80X branch. I've added it back in for consistency, but should we just scrap it?
- the code now runs ALL IDs for ALL years (even though they may not be applicable). Is it better to explicitly **not** run the "wrong" IDs for a given year and just store them as 0s? It doesn't really save space or time, more of a conceptual question

### DIFF
--- a/core/include/Electron.h
+++ b/core/include/Electron.h
@@ -58,13 +58,13 @@ class Electron : public RecParticle {
 
   Electron(){
 
-    m_supercluster_eta = 0; 
-    m_supercluster_phi = 0; 
-    m_dB = 0; 
-    m_neutralHadronIso = 0; 
-    m_chargedHadronIso = 0; 
+    m_supercluster_eta = 0;
+    m_supercluster_phi = 0;
+    m_dB = 0;
+    m_neutralHadronIso = 0;
+    m_chargedHadronIso = 0;
     m_photonIso = 0;
-    m_trackIso = 0; 
+    m_trackIso = 0;
     m_puChargedHadronIso = 0;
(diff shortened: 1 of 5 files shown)


raggleton:
Hmm failed because of the plotting comparison script, probably because I'm adding in new methods, which the plotter can't handle. Will try to fix and then rerun.

raggleton:
OK, so it seems that the way I do the plotting at the moment cannot handle changes in the class methods... need to rethink a bit how the plotting happens, not a quick fix (it's non-trivial because you're trying to inspect data structures that ROOT hides from you, so you have to rely on the class dictionaries, which depends on the class in the release _not_ in the ntuple...).

I'm going to leave this open for now unless anyone needs it urgently, since it's a good testing ground for this sort of automated plotting, plus there are some outstanding questions

PR #1086, opened 2019-01-11
update JetID for AK4CHS and create the new ID for AK4PUPPI

anmalara:


raggleton:
Reference to the recommendations: https://twiki.cern.ch/twiki/bin/view/CMS/JetID13TeVRun2017?rev=6

Will need porting to all the other 94X & 102X branches (although for 102X it gets more complicated as need to handle 2016 & 2017 IDs...)

raggleton:
Should also add that for the PUPPI IDs, where it states "charged multiplicity" on the twiki page, we have used (puppiMultiplicity - neutralPuppiMultiplicity).  Think this is OK, need to check with Andreas

PR #1087, opened 2019-01-11
enable storage of all partons for herwig

karavdin:


raggleton:
For the XML files, should they be in the RunII_94X_v3 directory instead?

karavdin:
> For the XML files, should they be in the RunII_94X_v3 directory instead?

Done!

PR #1088, opened 2019-01-14
added the pileupjet ID (only available for AK4 CHS jets)

abenecke:


raggleton:
Pipeline will probably fail as it changes the class members, which I haven't got around to plotting correctly.

Looks fine otherwise.

raggleton:
@ahinzmann does it make sense to also run this on AK8 CHS jets?

raggleton:
Some of the plotting did work, here is 2016v2 single mu data. Some interesting things:

raggleton:
Ok, so plotting pileupID vs eta shows exactly where those peaks crop up. -2 is for eta > 5 (producer only setup for eta < 5, see https://github.com/cms-sw/cmssw/blob/CMSSW_10_2_6/RecoJets/JetProducers/python/PileupJetIDParams_cfi.py), middle peak at ~-0.5 is for forward region.

slimmedjetspileupid_vs_slimmedjetseta

raggleton:
@abenecke since 0 is a valid ID value, can you change the default in the Jet constructor (https://github.com/UHH2/UHH2/blob/7688c8ef7a00944679fb7884860ce2ce61de1605/core/include/Jet.h#L45) to -2 instead? Once that is done it can be merged, thanks

PR #1091, opened 2019-01-21
[102X] Move to CMSSW_10_2_10 + properly do L1 prefiring

raggleton:
With 10_2_10, they integrated the L1 ECAL prefiring module properly: https://github.com/cms-sw/cmssw/releases/tag/CMSSW_10_2_10.

This PR now changes the install script to use 10_2_10, and removes the L1 prefire merge-topic. We also now use this new module, which thankfully removes all the faff with finding the weights ROOT file. The user no longer needs to specify anything in their CRAB config, nor worry about changing paths in their `ntuple_generator.py`.

Note that the label capitalisation changed, hence the changes in `NtupleWriter.cc`

(I'm also tempted to just import the cff file and save a few lines of code? Although more explicit here what's going on (e.g. systematics size))

### DIFF
--- a/core/plugins/NtupleWriter.cc
+++ b/core/plugins/NtupleWriter.cc
@@ -559,9 +559,9 @@ NtupleWriter::NtupleWriter(const edm::ParameterSet& iConfig): outfile(0), tr(0),
   branch(tr, "prefiringWeightDown", &event->prefiringWeightDown);
   if(doPrefire){
     std::string prefire_source = iConfig.getParameter<std::string>("prefire_source");
-    prefweight_token = consumes<double>(edm::InputTag(prefire_source, "NonPrefiringProb"));
-    prefweightup_token = consumes<double>(edm::InputTag(prefire_source, "NonPrefiringProbUp"));
-    prefweightdown_token = consumes<double>(edm::InputTag(prefire_source, "NonPrefiringProbDown"));
+    prefweight_token = consumes<double>(edm::InputTag(prefire_source, "nonPrefiringProb"));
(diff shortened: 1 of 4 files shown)


raggleton:
Failure was during `git cms-init`:
```
Fetching origin
fatal: unable to access 'https://github.com/cms-sw/cmssw.git/': The requested URL returned error: 503
error: Could not fetch origin
```

will retry job & see if it works, bit unusual

raggleton:
OK, something weird has happened with the prefire up/down weights, they're different than in 10_2_6, e.g. https://uhh2-integration.web.cern.ch/UHH2integration/test1091/mc_2016v2_TTbar_figs/prefiringWeightUp.pdf

prefiringweightup

I don't know why - it should be setup exactly the same way (same 0.2 factor). Either something in the ROOT file changed, or it's being used differently subtely?

raggleton:
OK I think I now know why. In the reference code, the up/down variations were computed as the larger/smaller of the stat uncertainty (from ROOT map) and the syst uncert (1±0.2). In the new code, the variation is both of those combined in quadrature.

See:

https://github.com/lathomas/cmssw/blob/L1Prefiring_10_2_6/L1Prefiring/EventWeightProducer/plugins/L1ECALPrefiringWeightProducer.cc#L198..L199

vs

https://github.com/cms-sw/cmssw/blob/CMSSW_10_2_10/PhysicsTools/PatUtils/plugins/L1ECALPrefiringWeightProducer.cc#L201..L202

I haven't `cout`'d explicitly to test this, but given that the central prefire weight remains the same, this looks like the cause.

PR #1092, opened 2019-01-22
[102X] Save year arg in `generate_process()` to Event class

raggleton:
This saves the `year` parameter used in the ntuple_generator to the Ntuple tree itself.

Analysis modules can then dynamically change their behaviour depending on this parameter, or the user can choose how to utilise it.

A string was chosen instead of an int, because the latter relies on some immutable mapping (or rather, never changing the order of existing values). This way, the user can immediately inspect and has more control over how to react. 1 string is also a very small overhead.

I guess the only question is: is "year" a good variable name? Or should it be something like "era"? (Can change the arg name in `generate_process()` as well)

### DIFF
--- a/core/include/Event.h
+++ b/core/include/Event.h
@@ -18,6 +18,7 @@ public:
   int run;
   int luminosityBlock;
   long long event;
+  std::string year;
 
   float rho;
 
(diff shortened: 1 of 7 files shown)


raggleton:
OK, can see new variable there in the plots, with a negligible increase in filesize.

(One slight annoyance is that the y looks weird for the new variable, but opening the ROOT file it is fine)

PR #1093, opened 2019-01-22
[102X] Store MET with EE noise mitigation only for 2017

raggleton:
Same as in 94X_v3. Note that the necessary updates went into 10_2_7 release, so no extra recipes steps required since we are now on 10_2_10 (see #1091 )

References: 
https://twiki.cern.ch/twiki/bin/viewauth/CMS/MissingETUncertaintyPrescription?rev=89#Instructions_for_9_4_X_X_9_or_10
https://github.com/cms-sw/cmssw/releases/tag/CMSSW_10_2_7

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -238,6 +238,26 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=150.):
 
 
     ###############################################
+    # Modified TypeI MET
+    #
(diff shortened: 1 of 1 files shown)


raggleton:
Can see new MET collection only in 2017 samples, is different to slimmedMET, all uncerts filled, x & y components not the same, uncorr not same as corr. Only 0s are for E & z components.
Adds ~2% to total filesize, nothing statistically significant to runtime.

PR #1095, opened 2019-01-22
[102X] simplify  charged/neutral component calculation

karavdin:
Simplify the code for charged/neutral component calculation to make it general and generator independent.

### DIFF
--- a/core/plugins/NtupleWriter.cc
+++ b/core/plugins/NtupleWriter.cc
@@ -1024,32 +1024,17 @@ bool NtupleWriter::filter(edm::Event& iEvent, const edm::EventSetup& iSetup) {
          double nhf = 0;
          double nef = 0;
          for (unsigned int k = 0; k < daughters.size(); k++) {
-           switch(abs(daughters[k]->pdgId())){
-             case 11: //electron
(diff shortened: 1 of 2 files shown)


core/plugins/NtupleWriter.cc
```diff
@@ -1024,32 +1024,17 @@ bool NtupleWriter::filter(edm::Event& iEvent, const edm::EventSetup& iSetup) {
...
-               nhf += daughters[k]->energy();
-               break;
-             case 22: //photon
-               nef += daughters[k]->energy();
-               break;
-           }
-         }
+	   if(abs(daughters[k]->pdgId())==11) 
+	     cef += daughters[k]->energy();
+	   if(abs(daughters[k]->pdgId())==22)
```
raggleton: Should this be an `else if`? Otherwise an electron will also contribute to `chf`

raggleton:
While you're revamping this, do we also want a muon energy fraction like for reco jets?

raggleton:
chf & cef in gentopjets same as before. now see non-0 muef in gentopjets. 

(for some reason it comes under "diff means" rather than "added hists" as it should...)

PR #1096, opened 2019-01-23
[102X] MET fixes: rawCHS for 2016v2, add sumEt, use proper MET significance

raggleton:
A hodgepodge of MET changes:

- for 2016v2, rawCHS is not stored. The code does run & store something, but no idea where it gets those values from. This variable often shows up in our tests as differing (with massive values sometimes), even though nothing has changed. So now I've just it to the usual MET px & py.

- now store sumET in MET

- replaced `MET::mEtSig` (which was actually = MET/sqrt(sumEt)) with `MET::mEtSignificance`, taken from the full covariance matrix. I changed the name (a) to match the `pat::MET` method, and (b) to ensure that any analyses using `mEtSig` are alerted to this change. The old `mEtSig` is still available, since we now store sumEt.
Note that for gen met, the significance matrix isn't stored, and instead requires explicit calculation. Given we don't even store gen met (should we?) I have not enacted this, but left a comment incase of future development there.

### DIFF
--- a/core/include/MET.h
+++ b/core/include/MET.h
@@ -6,6 +6,7 @@ public:
   MET(){
       m_pt = 0;
       m_phi = 0;
+      m_sumEt = 0;
       m_mEtSig = 0;
       m_uncorr_pt = 0;
       m_uncorr_phi = 0;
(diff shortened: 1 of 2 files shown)


karavdin:
GEN MET would be very useful, even without significance matrix.

raggleton:
OK, will merge this PR, then do a separate one for GenMet.

In this PR we can see values being filled in all MET collections, across all years, looks good to me

PR #1101, opened 2019-01-24
[102X] introduce GenJet class

karavdin:
proper GenJet class for generated jets with gen.particles, flavor and energy fractions; Flavor information is copied from "slimmedGenJetsFlavourInfos".
Not needed any more GenJetWithParts is removed. 
To be done: proper linking for mother and daughters genparticles in NtupleWriter::add_genpart.


raggleton:
cmsRun on 2016v2 MC failed, since it doesn't have any `slimmedGenJetsFlavourInfos` collection. Looks like we'll have to make it ourselves. I'll try and find a snippet to create it.

That has also led me to spot another thing: we get the flavour info from `slimmedGenJetsFlavourInfos`, but is that only for AK4 GenJets? Or do they pack both AK4 & AK8 flavour info into the one collection? There isn't another `reco::JetFlavourInfoMatchingCollection` in MINIAODSIM so either they pack, or there's no info for AK8 GenJets... (ditto for the gensubjets)

I also wonder what to do if someone wants another GenJet collection in, how to get the flavour info. It would probably require each genjet collection to be a `cms.PSet()` with a source, and a flavour collection source.

core/include/GenJet.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include "Particle.h"
+
+
+class GenJet : public Particle{
```
raggleton: Do we want to inherit from `FlavorParticle` instead for parity with the `Jet` class? I realise though that it isn't perfect, since like in the `Jet` class you'd have to add in a separate getter/setter for hadron flavour.

karavdin:
Concerning AK8 GenJets, I think it mostly used for top jets, for which hadronFlavor definition does not make sense and partonFlavor can be obtained by matching to ME parton. Therefore I am not sure we need monster from CMSSSW for fat Gen jets.

raggleton:
For 2016v2 MC, this is the snippet you need to add:

```
if year == "2016v2":
    process.load("PhysicsTools.JetMCAlgos.HadronAndPartonSelector_cfi")
    process.selectedHadronsAndPartonsForGenJetsFlavourInfos.particles = "prunedGenParticles"
    task.add(process.selectedHadronsAndPartonsForGenJetsFlavourInfos)

    from PhysicsTools.JetMCAlgos.AK4GenJetFlavourInfos_cfi import ak4GenJetFlavourInfos
    process.slimmedGenJetsFlavourInfos = ak4GenJetFlavourInfos.clone(jets="slimmedGenJets")
    task.add(process.slimmedGenJetsFlavourInfos)
```

this will make a `slimmedGenJetsFlavourInfos` that Ntuplewriter can then read. 
Bonus: it will recalculate the parton flavour using a definition that's consistent with that used in 94X (2017) & onwards - i.e. ignoring status codes, so it will work for not only Pythia8 but Herwig++ etc.

raggleton:
cmsRun jobs run fine now, failure due to trying to plot the different types. Deployed a fix, will rerun pipeline

raggleton:
Couple of things:

- now using British "flavour" instead of "flavor" for the getters/setters (although still using `FlavorParticle`)
- `Jet` already had a `hadronFlavor()` method, so now that it inherits from `FlavorParticle`, it has both `hadronFlavour()` and `hadronFlavor()`, we should only have one
- `ak8GenJetsSoftDrop` has lost its energy fractions, they're all 0 now, not clear why
- Looking at the genParticle_indices, it looks like nearly all of the genparticles get linked to a jet (not massively surprising), so I don't think there's any obvious way of slimming down the # of genparticles we store

karavdin:
New setters/getters copy values from MINIAOD and I kept the same names with British "flavour". 
For GenParticles the ideal solution would be to store them linked to jets if flag (smth like "doAllGenParticlesPythia8") is enabled. GenParticles flags need some cleaning, plan to do it in the next iteration.

PR #1099, opened 2019-01-24
added DeepFlavour, DeepBoostedAK8 and DeepDoubleX btagging discriminants

abenecke:
added deepflavour, deepBoosted, deepDoubleB discriminants

PR #1102, opened 2019-01-28
[102X] Retire HepTopTagger & CA15

raggleton:
Checked with Roman & Gregor, there are no plans to carry on HepTopTagger studies. So it's time to retire this collection & all the associated CA15 collections. I've merely commented them out, incase one needs them in future - they were tricky to enable in the first place.

This should save us a good chunk of processing time & some disk space, I hope


raggleton:
So this saves us about 1KB/event, more for jet-ty datasets, and about 0.1s / event (timings vary a bit, but all point towards that order of saving). Pleasingly, these savings are a tad more than added by the deepflavour additions in #1099

PR #1104, opened 2019-01-29
[102X] Electron::fbrem now stores saner default value

raggleton:
Default for fraction attributed to Bremsstrahlung in miniaod is -1e-30, which makes automated plotting difficult. This sets the minimum to -100 instead, which should be a safe, obvious value, and still show some differentiation power of 0 - 1  vs ==0.

(Aside, I see some electrons with -ve fbrem, does that even make sense? The code to calculate this is: https://github.com/cms-sw/cmssw/blob/CMSSW_10_2_10/RecoEgamma/EgammaElectronAlgos/src/GsfElectronAlgo.cc#L1482 , essentially `fbrem = (inMomentum - outMomentum) / inMomentum`, so -ve values indicate `outMomentum > inMomentum`...)

### DIFF
--- a/core/plugins/NtupleWriterLeptons.cxx
+++ b/core/plugins/NtupleWriterLeptons.cxx
@@ -1,3 +1,5 @@
+#include <algorithm>
+
 #include "UHH2/core/plugins/NtupleWriterLeptons.h"
 #include "UHH2/core/include/AnalysisModule.h"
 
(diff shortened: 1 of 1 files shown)


raggleton:
For some reason, most of the websites deployed but aren't actually there on the webserver... however the change is visible on the plots (under Very Large Range)

PR #1106, opened 2019-01-29
[102X] fix energy fractions for gen.topjets, flag for genjet constituents

karavdin:
fill gen energy fractions for self-clustered gentopjet, add flag to switch storing of genjet constituents on/off


raggleton:
Do we want `doGenJetConstituents` as a all-or-nothing bool, or a user-defined int? Latter could help with trimming disk space if one only wants the constituents for the leading 2 or 3 jets.

PR #1107, opened 2019-01-29
[102X] Change default value of m_genjet_index, add getter for genjet_index

raggleton:
Changed to -1 as 0 is a valid index

### DIFF
--- a/core/include/Jet.h
+++ b/core/include/Jet.h
@@ -40,7 +40,7 @@ class Jet : public FlavorParticle {
     m_btag_BoostedDoubleSecondaryVertexCA15 = 0;
     m_JEC_factor_raw = 0;
     m_JEC_L1factor_raw = 0;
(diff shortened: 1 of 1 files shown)


raggleton:
Useless without a getter as well, adding one and redoing pipeline

karavdin:
Hists don't look like desired change:
https://uhh2-integration.web.cern.ch/UHH2integration/test1107/mc_2017_TTHadronic_figs/slimmedJets.genjet_index.pdf
Is default overwritten somewhere?

raggleton:
Urgh somehow my 2nd commit changed it back to 0. Will fix and rerun tests

raggleton:
Finally, looks like they're all set at -1

(Side note: they count as "Added Hists" and not "Diff" since there was no getter before, which is how the automated plotter works)

PR #1111, opened 2019-01-29
Run ii 94 x v3

anmalara:
As far as I could, I tried to import the significant changes.
Probably there could be further updates.

The main changes are for:
Electron ID
Pu root files
Pu reweight module


raggleton:
Could you remove JECDatabase submodule? I'm hoping having add & remove in the history doesn't bloat the git tree size...

Also ideally one cherry-picks the commits across to retain the proper authorage, provenance, & github comments etc, like is done for other branches (although I realise not all commits are nice and atomic). It can be a faff, this is the reason why I fall behind on doing it...

anmalara:
Haven't notice I add to the cache. Sorry about that

PR #1112, opened 2019-01-31
[102_X] Proper pat jets for RECO xcone and hotvr

karavdin:
Create pat jets for reco XCONE and HOTVR in similar way like it's done for genjets. Also linking to constituents is fixed for reco and gen XCONE and HOTVR jets.

### DIFF
--- a/core/plugins/GenHOTVRProducer.cc
+++ b/core/plugins/GenHOTVRProducer.cc
@@ -139,7 +139,6 @@ GenHOTVRProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup)
   iEvent.getByToken(input_vertex_token_ , pvCollection);
   if (!pvCollection->empty()) vertex_=pvCollection->begin()->position();
   else  vertex_=reco::Jet::Point(0,0,0);
-  //  cout<<"vertex_ = "<<vertex_.x()<<" "<<vertex_.y()<<" "<<vertex_.z()<<endl;
   particles_.clear(); 
 
   edm::Handle<edm::View<reco::Candidate>> particles;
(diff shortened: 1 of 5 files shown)


raggleton:
Excellent, we now have EFs, multiplicities for both the fat jets and their subjets. (This is something I never tackled as it looks too scary 🙀 )

One thing to note is that for some samples, there are slightly fewer XCone subjets than before. From the plots it looks like its dumping the 0pt ones. Ah I see that in https://github.com/UHH2/UHH2/blob/546a4460303cadda84c703caf8f044cf22ea4785/core/plugins/XConeProducer.cc#L289 we actually fill the `subjetCollection`, which later gets put into the event. So the adding of null subjets doesn't actually get propagated into the event. Would you mind just deleting the lines https://github.com/UHH2/UHH2/blob/546a4460303cadda84c703caf8f044cf22ea4785/core/plugins/XConeProducer.cc#L297..L301 for consistency?

raggleton:
Also a reminder for future me that this also convert the RECO producers from Global to Stream producers (the GEN ones were already - I dunno why I changed the RECO ones to globals)

raggleton:
Also just had a thought - do we also want the PUPPI XCone & HOTVR to have their Puppi multiplicities filled as well @rkogler @ denschwarz? (Not for this PR, just a future possibility)

Actually that may not even be possible, since the input collection is `reco::Candidate` to handle both CHS and PUPPI, but PUPPI-specific methods would need to read in `PackedCandidate` objects (see change in https://github.com/UHH2/UHH2/commit/6613cae77ccbed8c2bf14c251e5fe38cc4fd52c4).

karavdin:
> Excellent, we now have EFs, multiplicities for both the fat jets and their subjets. (This is something I never tackled as it looks too scary scream_cat )
> 
> One thing to note is that for some samples, there are slightly fewer XCone subjets than before. From the plots it looks like its dumping the 0pt ones. Ah I see that in
> 
> [UHH2/core/plugins/XConeProducer.cc](https://github.com/UHH2/UHH2/blob/546a4460303cadda84c703caf8f044cf22ea4785/core/plugins/XConeProducer.cc#L289)
> 
> Line 289 in [546a446](/UHH2/UHH2/commit/546a4460303cadda84c703caf8f044cf22ea4785)
>  subjetCollection->push_back(subjet); 
> we actually fill the `subjetCollection`, which later gets put into the event. So the adding of null subjets doesn't actually get propagated into the event. Would you mind just deleting the lines https://github.com/UHH2/UHH2/blob/546a4460303cadda84c703caf8f044cf22ea4785/core/plugins/XConeProducer.cc#L297..L301 for consistency?

Empty subjets are removed as suggested

PR #1113, opened 2019-02-01
[102X] add XCONE for dijet (JERC) studies

karavdin:
Number of xcone collections for dijet/JERC related studies. Have to be produced only for QCD MC and JetHT DATA, for the rest can be switched off.


raggleton:
FYI `NSubJets = cms.uint32(1)` will still cluster 1 subjet - is this the desired effect? Subjet clustering is only disabled for `NSubJets = cms.uint32(0)` (I should really write this down in the config)

karavdin:
> FYI `NSubJets = cms.uint32(1)` will still cluster 1 subjet - is this the desired effect? Subjet clustering is only disabled for `NSubJets = cms.uint32(0)` (I should really write this down in the config)

Yes, I'd like to have something like "hard core" of the jet to check how often half of jet cone contains >90% of jet energy.

PR #1116, opened 2019-02-04
[102X] Store pileup working points

raggleton:
Add in `Jet` tags for the loose/medium/tight pileup working points. Also add in example module showing how to use, and histogramming frequency of each ID to test.

### DIFF
--- a/core/include/Jet.h
+++ b/core/include/Jet.h
@@ -9,7 +9,19 @@
 class Jet : public FlavorParticle {
 
  public:
-  enum tag { /* for future use (more b-taggers, etc.) */ };
+  enum tag {
+    // Pileup ID working points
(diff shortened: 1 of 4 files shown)


raggleton:
Forgot to say, that I figured this would be the most consistent way of dealing with loose/medium/tight WPs, like is done for `Electron`s. Can't think of a significantly better way, although it does mean they don't appear in the automated plotting (hence the example module instead).

This also negates the need for us to implement an additional module to calculate the pass/fail ourselves from the raw pileup discriminant. We should only have one way of accessing pass/fail for each pileup WP (unlike for Electrons where we have both, and the module requires manual updating & bug fixing). Obviously this may change in future depending on what happens with pileupID development.

PR #1117, opened 2019-02-06
[102X] Recluster AK8PUPPI for TopJets ourselves

raggleton:
Several things happen here:

- We now recluster our own AK8 PUPPI jets with all the melange of substructure and tagging. This replaces our reliance on slimmedJetsAK8, ensuring we have a consistent and flexible collection across all datasets, at the cost of extra runtime. It also means we don't have to worry about daughters really being subjets in some datasets. The new AK8 PUPPI mirrors the already existing AK8 CHS used for TopJets, with the extra deep* stuff

- Sorted out the DeepFlavour/Tagging + PUPPI PATPuppiJetSpecificProducer modules so that we only have one call to `updateJetCollection`, removing the extra jet collections. The trick is to do the PUPPI bit only for the very last `PATJetUpdater` module (since we use the collection name for userFloats) [Aside: the deepFlavour stuff cannot be added in with `addJetCollection`, only `updateJetCollection`]

- Added a crucial but totally undocumented fix to ensure the DeepBoostedJetTags are calculated properly for reclustered PUPPI jets

Minor stuff:

- Update the JEC levels in `add_fatjets_subjets` to ignore L1FastJet if PUPPI, and add L23Residual for data. Shouldn't be important as the user then applies JECs, but I thought I'd do it while tidying up.

- Lots of formatting & whitespace changes, plus more comments for future selves. Would recommend ignoring whitespace changes when looking at this PR diff (Diff settings in upper right -> Hide whitespace changes -> Apply)

### DIFF
--- a/core/plugins/NtupleWriterJets.cxx
+++ b/core/plugins/NtupleWriterJets.cxx
@@ -42,15 +42,8 @@ float getPatJetUserFloat(const pat::Jet & jet, const std::string & key, float de
 
 // Generate the name of the puppiJetSpecificProducer module
 // So ugly, really should get the user to configure this in the NtupleWriter py
-std::string getPuppiJetSpecificProducer(const std::string & name){
+std::string getPuppiJetSpecificProducer(const std::string & name) {
   std::string multiplicity_name = "patPuppiJetSpecificProducer"+name;
-  if(multiplicity_name == "patPuppiJetSpecificProducerselectedUpdatedPatJetsSlimmedJetsPuppiNewDFTraining"){
-    multiplicity_name = "patPuppiJetSpecificProducerupdatedPatJetsTransientCorrectedSlimmedJetsPuppiNewDFTraining";
-  }else if(multiplicity_name == "patPuppiJetSpecificProducerselectedUpdatedPatJetsPatJetsAK8PFPUPPINewDFTraining"){
-   multiplicity_name = "patPuppiJetSpecificProducerupdatedPatJetsTransientCorrectedPatJetsAK8PFPUPPINewDFTraining";
-  }else if("selectedUpdatedPatJetsSlimmedJetsAK8NewDFTraining"){
-    multiplicity_name = "patPuppiJetSpecificProducerupdatedPatJetsTransientCorrectedSlimmedJetsAK8NewDFTraining";
-  }
   return multiplicity_name;
 }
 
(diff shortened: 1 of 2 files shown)


raggleton:
Looking at how they do it in CMSSW (https://github.com/cms-sw/cmssw/blob/CMSSW_10_2_10/PhysicsTools/PatAlgos/python/slimming/applyDeepBtagging_cff.py), it appears that they apply the `pfDeepFlavourJetTags` to slimmedJets (i.e. AK4 CHS), which we don't currently.

There are also some extra AK8 ones that they add in that we don't:

-'pfDeepBoostedDiscriminatorsJetTags:H4qvsQCD'
-'pfDeepBoostedDiscriminatorsJetTags:HbbvsQCD'
-'pfDeepBoostedDiscriminatorsJetTags:TvsQCD'
-'pfDeepBoostedDiscriminatorsJetTags:WvsQCD'
-'pfDeepBoostedDiscriminatorsJetTags:ZbbvsQCD'
-'pfDeepBoostedDiscriminatorsJetTags:ZvsQCD'
-'pfDeepDoubleBvLJetTags:probHbb'
-'pfDeepDoubleBvLJetTags:probQCD'
-'pfDeepDoubleCvBJetTags:probHbb'
-'pfDeepDoubleCvBJetTags:probHcc'
-'pfDeepDoubleCvLJetTags:probHcc'
-'pfDeepDoubleCvLJetTags:probQCD'
-'pfMassDecorrelatedDeepBoostedDiscriminatorsJetTags:H4qvsQCD'
-'pfMassDecorrelatedDeepBoostedDiscriminatorsJetTags:HbbvsQCD'
-'pfMassDecorrelatedDeepBoostedDiscriminatorsJetTags:ZbbvsQCD'
-'pfMassDecorrelatedDeepBoostedDiscriminatorsJetTags:ZvsQCD'
-'pfMassIndependentDeepDoubleBvLJetTags:probHbb'
-'pfMassIndependentDeepDoubleBvLJetTags:probQCD'
-'pfMassIndependentDeepDoubleCvBJetTags:probHbb'
-'pfMassIndependentDeepDoubleCvBJetTags:probHcc'
-'pfMassIndependentDeepDoubleCvLJetTags:probHcc'
-'pfMassIndependentDeepDoubleCvLJetTags:probQCD'

Are we OK without these?

raggleton:
Hmm  the differences in `packedPatJetsAk8CHSJets_SoftDropCHS` are all at low pT.  I think this is because previously we were using the jets made with a low pT cut (10 GeV) for that, then applied a cut of 150 when storing in ntuples. However now there is a cut on the FastjetJetProducer already at 150, then JECs are applied, then the standard 150 cut when storing in ntuples. I don't quite understand how, but it seems we get more jets doing it with the higher cut on the FastjetJetProducer...

karavdin:
@raggleton, could you post example plot, please? TTbarHadronic looks fine to me:
https://uhh2-integration.web.cern.ch/UHH2integration/test1117/mc_2018_TTHadronic_figs/packedPatJetsAk8CHSJets_SoftDropCHS.pt.pdf

raggleton:
Ah OK I failed to read my own legend, got it the wrong way round. With this PR one loses a few of the lowest pT jets, as expected

PR #1118, opened 2019-02-06
[102X] add storage of PF candidates for selected number of reconstructed jets

karavdin:
The feature is enabled by "doPFJetConstituents" flag (=number of jets for which PF candidates should be stored and linked to Jet).  At the moment one flag for all jet collections for simplicity. Might be useful for comparing leading jets between each other.


karavdin:
Attempt to fix storage of PF candidates for sub-jets. Does not work for _Softdrop collections, because constituencies are sub-jets and not PF candidates.

PR #1119, opened 2019-02-06
[102X] Set btag discriminant defaults to impossible value

raggleton:
0 is dangerous as it's a valid discriminant output. -2 should be outside the range of any discriminant, but still makes automated plotting useful


raggleton:
OK thankfully looks like only null btag discriminants were the changed hists.

PR #1120, opened 2019-02-06
[102X] Better Jet/TopJet collection names

raggleton:
Now we can relatively easily rename our jet collections by using new function `rename_module`. This can also handle userData (e.g. for `PATJetUpdater` modules) so that e.g. PUPPI multiplicities are not broken (because we use the collection name to lookup the userFloat)

All AK{4,8}{CHS,PUPPI} collections have been renamed, except `slimmedJets` since we don't touch it. But we can rename that as well to `jetsAk4CHS` for consistency.

So we now have:
Jets: `slimmedJets`, `jetsAk4Puppi`, `jetsAk8CHS`, `jetsAk8Puppi`  (the latter 2 are the low-pT ones for JERC)
TopJets: `jetsAk8CHSSubstructure_softDropCHS`, `jetsAk8PuppiSubstructure_softDropPUPPI`

Are the new names OK? Or do we also want something in them to reflect whether we have reclustered them?

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -1105,8 +1105,80 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=150.):
             '%s:HFEMPuppiMultiplicity' % puppi_mult_name
         ]
 
-    # Higgs tagging commissioning
+    def rename_module(process, task, current_name, new_name, update_userData=True):
+        """Rename a module in a process, with option to also update userData.
+        This clones the existing module at process.current_name into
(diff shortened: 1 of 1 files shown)


karavdin:
This should make our life simple. Thanks a lot!
I like the proposed names, adding more info in them can kill the beauty of (relatively) short-names. Concerning the reclustering and other important details I think we should create wiki-page, which people can consult instead of going through the code.

raggleton:
Yeah, I was thinking of having the table on the corresponding NTuple instruction page, so that it's all kept together for that particular branch

PR #1121, opened 2019-02-06
[102X]  add CHS MET for 2016 data

karavdin:
Adding MET collection ("slMETsCHS") with explicit clustering of CHS MET for 2016 data and MC, for which chs MET user-floats were not available


raggleton:
I'll kill the current testing pipeline and restart it

PR #1127, opened 2019-02-11
[102X] add DeepBoosted info to AK8CHS jets

karavdin:
Add DeepBoosted info to AK8CHS jets. Also DeepFlavor info is filled for AK4 jets, including AK4 sub-jets of top-jets.

### DIFF
--- a/core/include/Jet.h
+++ b/core/include/Jet.h
@@ -50,80 +50,13 @@ class Jet : public FlavorParticle {
     m_btag_combinedSecondaryVertexMVA = -2;
     m_btag_DeepCSV_probb = -2;
     m_btag_DeepCSV_probbb = -2;
-    m_btag_BoostedDoubleSecondaryVertexAK8 = -2;
-    m_btag_BoostedDoubleSecondaryVertexCA15 = -2;
     m_btag_DeepFlavour_probbb=-2;
     m_btag_DeepFlavour_probb=-2;
     m_btag_DeepFlavour_problepb=-2;
     m_btag_DeepFlavour_probc=-2;
     m_btag_DeepFlavour_probuds=-2;
     m_btag_DeepFlavour_probg=-2;
-    m_btag_MassDecorrelatedDeepBoosted_bbvsLight=-2;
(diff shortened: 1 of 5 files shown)


raggleton:
Hmm somehow we've gained/lost some AK4 CHS jets, and some PFParticles - I suspect this is because we now run `updateJetCollection()` on slimmedJets, which will apply JECs in the Global Tag that are different to the ones used when creating the sample. This may shift the pT a bit, possibly leading to some jets failing the pT > 10 cut in the ntuplewriter?

Other distributions like `jetsAk4CHS.btag_DeepFlavour_b()` look **very** different - I guess the new distributions are correct, and whatever was stored in there before was junk? Didn't even know it was stored in there before

Bit worrying that it says `jetsAk8PuppiSubstructure_SoftDropPuppi.subjets().JEC_factor_raw()` plus other subjet distributions like energy fractions in mc_2016v2_TTbar have different means/RMSes - I can't see it in the plots, so either it may be a float-comparison issue rather than any genuine change. Will need to check the files explicitly.

karavdin:
Here is difference in number of events, extracted from https://github.com/UHH2/UHH2/pull/1127#issuecomment-462850115

|  | mc_2016v2_TTbar |  mc_2016v3_TTbar | mc_2017_TTHadronic | mc_2018_TTHadronic |
| ---- | ---- | ---- | ---- | ---- | 
| PF particles | 0 | -0.002% | +0.003% | 0 |
| jetAK4CHS | 0 | -2.84% | -4.98% | 0 | |
| comment |  | more events with jet pt<20GeV | more events with jet pt<20GeV |  |

Indeed the difference comes from low pt jets. Could it be that we are using different GT in respect to miniAOD production?

Concerning DeepFlavour for AK4CHS:
|  | mc_2016v2_TTbar |  mc_2016v3_TTbar | mc_2017_TTHadronic | mc_2018_TTHadronic |
| ---- | ---- | ---- | ---- | ---- | 
jetsAk4CHS.btag_DeepFlavour_* | not filled before/filled now | different | not filled before/filled now| the same| 

Maybe difference in mc_2016v3 related to absence of low pt jets?

PR #1128, opened 2019-02-12
[102X] Update Electron ID module to do electrons + photons, rename it

raggleton:
V.similar mechanism, might as well make it a general EGamma module instead. Have renamed files to reflect this.

Adds in hist of # passing for all photon IDs.


raggleton:
Hmm something weird is going on with the Puppi iso quantities in the photons, they seem to give different values each time they're run, which is worrying. Nothing to do with this PR (can see it also in 1127), but worth checking at some point

PR #1129, opened 2019-02-12
[102X] GenTopJet substructure vars

raggleton:
Store ungroomed & groomed gentopjet collections, both now have tau_1,2,3,4. Also add ECFs to GenTopJet class, only store for groomed genjets like for TopJets.
tau_2,3,4 seems to be calculated as 0 for groomed GenTopJet, I don't know why - the setup is the same as for TopJets, and tau_1 is non-zero.

Also tried to simplify sources a bit to avoid too much repetition.


raggleton:
Across all MC samples, tau2,3,4 for ak8GenJetsSoftDrop is 0. ECFs being filled, not sure if they look physically sensible though

PR #1130, opened 2019-02-12
[102X] add PUPPI multiplicities for AK8CHS jets (just in case)

karavdin:


raggleton:
Hmm so in theory for CHS, all PUPPI weights = 1? But if you compare the new puppiMultiplicity with numberOfDaughters (which should be the total multiplicity), they are very different:

https://uhh2-integration.web.cern.ch/UHH2integration/test1130/data_2016v2_JetHT_figs/jetsAk8CHSSubstructure_SoftDropCHS.puppiMultiplicity.pdf
https://uhh2-integration.web.cern.ch/UHH2integration/test1130/data_2016v2_JetHT_figs/jetsAk8CHSSubstructure_SoftDropCHS.numberOfDaughters.pdf

karavdin:
In my understanding, puppiMultiplicity and number of daughters can be different. @ahinzmann, @abenecke, it is correct?

ahinzmann:
The Puppi weight is a property of the packed candidates, thus when looking at the particles inside CHS jets, one can retrieve the PUPPI weight even though it is not used for the CHS jets, i.e. they are meaningless and rather reflect the PUPPI multiplictiy for a PUPPI jet containing similar set of particles

ahinzmann:
My take is we don't want to store puppi multiplicities for CHS jets, since they are meaningless

karavdin:
Thanks! In this case we better avoid calculation of it in the first place. Will do another PR with removing puppi multiplicities calculation for CHS jets.

PR #1131, opened 2019-02-13
[102X] Lower fatjet_ptmin to account for JECs

raggleton:
If JECs are generally > 1, then the effective cut on topjets is larger than `topjet_ptmin`. This way we retain jets from ~150 - 170 or so.


raggleton:
Now see more topjets in the lowest pt bins of the histograms, so hopefully cutting at something like 180-200 GeV is totally safe. Marginal increase in filesize, since it's only a few extra jets

PR #1136, opened 2019-02-13
[102X] add more flags for PF/GEN constituents storing

karavdin:
Upon request, handles to store PF and GEN constituents are changed. Now, in addition to storing constituents for certain number of jets (with flag doPFJetConstituentsNjets), one can also store them for (all) jets above pt threshold (with flag doPFJetConstituentsMinJetPt). If not used, min pt threshold should be set to 0 or negative value to be handled properly.
Each family of jets (AK4 jets, TopJets, HOTVR, XCone, XCone_dijet) has its own set of flags. The same is true for GEN jet families.

Minor changes: removed genXCone23 and added XCONE_dijet PUPPI jets.


raggleton:
FYI this one will take a while to go through the pipeline, it is now like 1.3s/event, plus more loads hists to copy across to the webserver

PR #1137, opened 2019-02-13
[102X] GenJet multiplicities

raggleton:
Add multiplicities to GenJet class, same as those for Jet class. Also done for GenTopJet explicitly from its subjets, like for the energy fractions.

Useful especially for the lepton genjets that require a cut on e.g.  ==1 electron and == 0 muon


raggleton:
Not quite sure why it's added the taus in my PR, even though my commits don't touch ntuple_generator.py... Looks like everything that is meant to be filled is filled (ak4GenJetsFat doesn't have any subjets, so they should be empty).

PR #1138, opened 2019-02-14
[102X] switch off collection/storage not needed by defualt 

karavdin:


raggleton:
Wow, so we save 2/3rds of disk space by turning off all those things... at 14kB /event, 10M events would be 140 GB, not too bad. And a reasonable bit of time saving as well.

BTW we will also have to turn off Dennis' lepton jets in #1139 once they have passed & been merged.

PR #1139, opened 2019-02-14
added Muon and Electron GenJets

denschwarz:
Two lists are created from stable GenParticles:

1. muons and photons that have a muon as a mother
2. electrons and photons that have electrons as a mother

These lists are then used as input for anti-kT clustering with R=0.1

Note: In Herwig++ the jets will not have photons included. But for the samples where this is actually used (2016 version of ttbar Powheg and aMC@NLO) it works.

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -67,7 +67,7 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=150.):
         met_sources_GL.extend(['slimmedMETsEGClean', 'slimmedMETsMuEGClean', 'slimmedMETsUncorrected'])
     if (year=="2016v2" or year=="2016v3"):
         met_sources_GL.extend(['slMETsCHS'])
-    
+
 
 
     bTagDiscriminators = [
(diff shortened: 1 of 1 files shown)


raggleton:
Urgh for some reason the webpages mostly didn't get uploaded to the server. Will see if I can do it manually, or just rerun whole pipeline again. Please don't merge for now

raggleton:
Now see nice distributions. Interestingly, there are a tiny fraction of events where the electronGenJet has 0 electron multiplicity - I guess the electron is soft and falls outside the ∆R < 0.1 cone? (same for muonGenJets)

BTW for reference:

- 2016v2 MC is powheg-pythia8 ttbar
- 2016v3 MC is powheg-pythia8 ttbar MC
- 2017 MC is amc@nloFXFX-pythia8 (ttleptonic) and powheg-pythia8 (tthadronic)
- 2018 MC is powheg-pythia8 for both tthadronic and ttleptonic

(all input files listed here: https://gitlab.cern.ch/raggleto/UHH2-integration/blob/master/scripts/cmsrun_jobs.py)

so we can see it works for the "main" generators

denschwarz:
Yes, when I tested I saw some jets where the energy fraction nef was exactly 1. So, these are events where only photons are clustered into the jet. But this is easy to clean on analysis level.

PR #1143, opened 2019-02-15
[102X] Turn off photon puppi isolation for 2016v2

raggleton:
This is because (a) the method didn't exist in 80X when the dataset was made, (b) the pat::Photon constructor doesn't initialise the puppiIsolation variables, so we just read junk values, hence we get values like 10^26, and that differ between runnings.


raggleton:
Now see photon PUPPI isolation values of == -1 for 2016v2 samples. All others unchanged (and don't have a massive range of values).

PR #1147, opened 2019-02-18
test signal samples for Z' analysis

karavdin:


raggleton:
Cancelled pipeline as clearly doesn't affect compilation or running.

BTW there are now 2 features to help avoid unnecessary pipeline running:

- if you include `[ci skip]` in your PR text contents, it will skip the entire testing chain
- if you include `[only compile]` in your PR text contents, it will only compile and skip the cmsRun jobs & website etc.

PR #1151, opened 2019-02-19
[102X] fill GenTopJet genparticles, add jet constituent example

raggleton:
- Fill GenTopJet constituents like for GenJets
- Remove unused bools for genjets, they don't appear to be used anywhere
- Setup PFParticles using `PFParticleCollection` name in XML like for other collections
- Add example module that reconstructs Jets, TopJets, GenJets, GenTopJets from their constituents, and tests & plots relative differences in pt/eta/phi

Interestingly, whilst the relative difference in pT is often small (< 1E-3), the relative difference in eta can be > 1E-2!
e.g. for `jetsAk4CHS` in MC 2017 (relative difference = |x-y| / max(x, y)):

jet_constit_pt_compare
image
image

### DIFF
--- a/core/plugins/NtupleWriter.cc
+++ b/core/plugins/NtupleWriter.cc
@@ -228,17 +228,11 @@ NtupleWriter::NtupleWriter(const edm::ParameterSet& iConfig): outfile(0), tr(0),
   doGenJetConstituentsNjets = iConfig.getParameter<unsigned>("doGenJetConstituentsNjets");
   doGenJetConstituentsMinJetPt = iConfig.getParameter<double>("doGenJetConstituentsMinJetPt");
   if(doGenJetConstituentsMinJetPt<1e-6) doGenJetConstituentsMinJetPt=2e4;
-  doGenJetConstituents = false;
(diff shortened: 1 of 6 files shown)


raggleton:
Hmm appears we have lost some final-state GenParticles, you can see it in https://uhh2-integration.web.cern.ch/UHH2integration/test1151/mc_2016v2_TTbar_figs/ak8GenJetsFat.genparticles_indices.pdf. I think what's happened is:

- previously, `doGenTopJetConstituentsMinJetPt` was set to -1. this then gets set automatically to 2e4
- that means that `doGenTopJetConstituents` became `true`
- that means `if(gentopjets[j].size()<doGenJetConstituents) add_genparts=true;` could feasibly evaluate `true`, since `true` is 1 when cast to a uint, so it would set `add_genparts=true` for the first gentopjet in the event
- `add_genparts` gets passed to `fill_geninfo_recojet()`, which goes and stores the genparticles

Now that the logic is fixed, it correctly doesn't store the genparticles (since we've turned all the options off)

PR #1154, opened 2019-02-19
[102X] electronID fixes

karavdin:
add MVA 2016 electronID tags; 
remove outdated electron IDs; 
update cut-based without iso, which might be useful for some analyses, e.g with boosted semileptonic top


raggleton:
I have no idea where the PR review table has gone... no error message in the logs to say it failed. Not crucial for this PR, since you wouldn't see any differences in the plots.

All webpages are here: https://uhh2-integration.web.cern.ch/UHH2integration/test1154/

PR #1155, opened 2019-02-19
[102X] fix issue with grid-control running

karavdin:
change a way how params from grid-control env are extracted
Now grid-control at least works with files stored locally at DESY
[ciskip]
[ci skip]


karavdin:
To save some time for future testers: 
"error code: 92" from ntuple_writer running through grid-control means it can't access the file. It seems nothing to do with ntuple_writer settings and related to the fact that grid-control doesn't fetch X509_USER_PROXY settings properly.

PR #1160, opened 2019-02-21
[102X] add stuff for advanced trigger studies

karavdin:
Store L1 triggers prescales, because if your trigger prescaled it's most probably done at L1. 
Also copied "doL1seed"  code from 94X_v3, which might be useful for L1 prefiring studies with 2017 data. Enabled for testing, should switched off by default in the production.

### DIFF
--- a/core/include/Event.h
+++ b/core/include/Event.h
@@ -44,7 +44,9 @@ public:
   std::vector< TopJet >* toppuppijets;
   MET* met;
   MET* genmet;
-    
+  std::vector< L1EGamma>* L1EG_seeds;
+  std::vector< L1Jet>* L1J_seeds;
+
   GenInfo* genInfo;
   std::vector< GenTopJet >* gentopjets;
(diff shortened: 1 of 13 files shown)


raggleton:
For future reference when I forget this in the future:

- a given HLT path can be seeded by multiple L1 seeds, each of which has a certain prescale. `triggerPrescalesL1min` and `triggerPrescalesL1max` refer to the minimum & maximum non-0 values across all those seed prescales. So if L1min = 1, there is at least one unprescaled seed. If L1min = L1max then either there is only 1 seed, or they all have the same prescale. Other cases are more complicated
- L1 prescale info not stored for 2016v2, only 2016v3 and later. Not clear if it should actually work or not
- `triggerPrescales`, `triggerPrescalesL1min`, `triggerPrescalesL1max` all have 1 appended when doing L1 seeds, to keep them consistent with `triggerNames_outbranch` and `triggerResults`.

PR #1162, opened 2019-02-21
[102X] Use NtupleWriter class for GenTopJets, do substructure+grooming properly

raggleton:
Taking inspiration from how TopJets are handled, introduce `NtupleWriterGenTopJet` class to be able to handle complex setup for GenTopJets.
It's complex because we may want grooming applied, storing subjets, but also calculating substructure & other quantities on the groomed jet (this requires 2 separate jet collections).
Crucially, the njettiness & ECF were not being calculated correctly previously, since they were being calculated on jets that only had subjets as their constituents. This is now fixed.
Now it is all tied together a bit more nicely, and provides more route for expansion if there are other variables to store.

We therefore have the following GenTopJet collections:
- ak8GenJetsFat: ungroomed, has tau_i, no ECF, no subjets
- ak8GenJetsSoftDropforsub: groomed, has tau_i, ECF, subjets

GenTopJets & their subjets have all their constituent info calculated & stored as well.

Also updated ExampleJetConstituents to also plot TopJet subjets and GenTopJet subjets.

PS I also want to change those collection names, but will do that in a later PR to keep things neater

### DIFF
--- a/core/plugins/NtupleWriter.cc
+++ b/core/plugins/NtupleWriter.cc
@@ -320,7 +320,7 @@ NtupleWriter::NtupleWriter(const edm::ParameterSet& iConfig): outfile(0), tr(0),
         writer_modules.emplace_back(new NtupleWriterElectrons(cfg, true, save_lepton_keys));
         //}
 
-}
+  }
   if(doPhotons){
       using uhh2::NtupleWriterPhotons;
       auto photon_source = iConfig.getParameter<edm::InputTag>("photon_sources");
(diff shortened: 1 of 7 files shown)


raggleton:
OK so can now see more sensible tau2/3/4. 
The constituent variables for the fatjet are actually the same, since before we did a sneaky trick of copying everything from the subjets instead. Good to see both ways are consistent.
The energy fractions were actually wrong beforehand, since they were summed directly. Now they make more sense.

PR #1167, opened 2019-02-26
[102X] Split 2017 config settings into 2017v1 & 2017v2

raggleton:
Since we need separate GTs & Eras for 2017v1 and v2, this adds extra parameters `2017v1` &`2017v2` for `generate_process()`, and retires `2017` argument.
Also add in corresponding configs. Purposely excluded data 2017v1 config since it is not really useful as people will be using the 31Mar18 dataset, which is counts as 2017v2 for us.

Updated bits in `generate_process()` that are conditional on year arg, e.g. MET EE fix.

So now:
- RunIIFall17MiniAODv2 MC and 31Mar2018 data should use `2017v2` configs
- RunIIFall17MiniAOD MC should use `2017v1` config

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -55,7 +55,7 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
         If the year argument is not one of the allowable options
     """
     year = str(year)  # sanitise string
-    acceptable_years = ["2016v2", "2016v3", "2017", "2018"]
+    acceptable_years = ["2016v2", "2016v3", "2017v1", "2017v2", "2018"]
     if year not in acceptable_years:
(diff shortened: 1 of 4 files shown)


raggleton:
Failed because it doesn't know how to handle the case of a config file missing in the reference branch... will fix and retry

raggleton:
Finally got it working. New entires for data_2017v2, mc_2017v1, and mc_2017v2 are in the table & plotted. (timing is exactly the same as it just copies timing data for comparison as no reference timing data)
Both v1 and v2 have modified MET, prefire weight, passEcalBadCalib bit.

Can't tell if the eras are correct yet, as there is no "before" comparison. Can merge this then run another PR to revert to builtin btag values for slimmedJets, etc.

PR #1170, opened 2019-02-27
[102X] Try the 2018 ReReco GT DO NOT MERGE

raggleton:
See if it works with 2018 RunD

I've also changed the 2018_singlemu file in the testing setup for a RunB ReReco file to also test if using the Prompt GT was doing Bad Things. The same one is added into the config here for local testing purposes.

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -283,7 +283,7 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
             "mc": "94X_mc2017_realistic_v17"
         },
         "2018": {
-            "data": "102X_dataRun2_Prompt_v6",
+            "data": "102X_dataRun2_Sep2018Rereco_v1",
             "mc": "102X_upgrade2018_realistic_v15",
         },
     }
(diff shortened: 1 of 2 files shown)


raggleton:
(I should point out that in the tables, "diff entries" is higher priority than "diff mean rms". Perhaps I should just merge them into one.)

So it looks like changing GT does affect both ReReco and Prompt in 2018. I suspect this is due to the JECs? (which here get recalculated for every jet collection). It looks like the changes are in the lowest jet pT bins, supporting this hypothesis.

Without comparing individual jets, it makes it v.difficult to see if there's an effect of the GTs on other variables like b tagging (I dunno if it's even affected by GT?)

karavdin:
JECs are different in GTs, as one can also see on JEC raw factor plots:
https://uhh2-integration.web.cern.ch/UHH2integration/test1170/data_2018_SingleMu_figs/jetsAk4CHS.JEC_factor_raw.pdf
More up to date GT might do better job for 2018, because it contains 2017 JEC, which should not be so far away from 2018 data as 2016 JEC. But stickly speaking both of JECs are wrong for 2018 data.

PR #1171, opened 2019-02-27
[102X] Revert to precalculated btag values for slimmedJets, slimmedJetsPuppi DO NOT MERGE

raggleton:
Following on from #1167, this reverts to the precalculated b tagging values for slimmedJets and slimmedJets PUPPI (ie jetsAk4CHS and jetsAk4Puppi) only for testing purposes. If we've done everything right with the eras, there should be no differences 🤞

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -1146,7 +1146,7 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
     # This MUST be run *After* JetSubstructurePacker, so that the subjets are already there,
     # otherwise the DeepBoostedJetTagInfoProducer will fail
     # Also add in PUPPI multiplicities while we're at it.
-    for name in ['slimmedJets', 'slimmedJetsPuppi', 'patJetsAK8PFPUPPI', 'packedPatJetsAk8PuppiJets','packedPatJetsAk8CHSJets']:
+    for name in ['slimmedJetsPuppi', 'patJetsAK8PFPUPPI', 'packedPatJetsAk8PuppiJets','packedPatJetsAk8CHSJets']:
         labelName = cap(name)
         is_ak8 = "ak8" in name.lower()
         is_puppi = "puppi" in name.lower()
@@ -1193,6 +1193,11 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
 #        discriminators = ak4btagDiscriminators[:]
         if is_ak8 and is_topjet:
             discriminators.extend(ak8btagDiscriminators)
+
+        if name == "slimmedJetsPuppi":
+            discriminators = None
+            updater_name = updater_src
+            updater_src = name
 
         updateJetCollection(
             process,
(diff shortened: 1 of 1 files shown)


raggleton:
I realise this test isn't as helpful as I thought, since it also removes any JECs done in the call to `updateJetCollection()`. I guess I either need to apply them to actually make this useful and avoid conflating JEC changes with btagging changes.

However, one can see from the `jetsAk4CHS.btag_DeepCSV()`

- v.similar in data_2016v3
- v.similar in data_2017v2
- v.similar in data_2018 (JetHT, Prompt RunD)
- fairly similar in data_2018 (SingleMu, ReReco RunB)

- v.similar in mc_2016v3
- v.similar in mc_2017v1 (biggest diff in null -2 bin)
- v.similar in mc_2017v2
- v.similar in mc_2018

At least none of the shapes look as radically different as we saw before using the incorrect eras, so it suggests we got it right?

(I still don't know why it's flagging some distributions as "different" when they look identical - both the means & RMSes are compared, and each is required to have a relative difference < 1E-9 between new & ref. Perhaps this tolerance is still too tight.)

raggleton:
Digging into the actual values, e.g. for the seemingly identical 2018 MC TTHadronic jetsAk4CHS.JEC_factor_raw(): https://uhh2-integration.web.cern.ch/UHH2integration/test1171/mc_2018_TTLeptonic_figs/jetsAk4CHS.JEC_factor_raw.pdf

So it turns out that there are 1367 jets (out of 7562) where the relative difference > 1E-9. Infact, the largest relative difference is 5.8e-07, and the smallest 6.1e-08. The majority are ~1e-7. 
All other 6195 jets have ==0 difference between the new & ref. I guess this is the difference between calculating JEC with full precision in AOD, and redoing it with MiniAOD (the GTs here is the same as used in production). To avoid this, I've raised the relative difference lower limit to 1E-6. That should ignore all these rounding errors, but still allow us to see very small differences.

raggleton:
So I'm now fairly satisfied that we have recalculated `btag_DeepCSV()` correctly for `jetsAk4CHS`, there is agreement in all samples (module some rounding differences).

However:

- how come the DeepFlavour distributions change in e.g. mc_2017v2: https://uhh2-integration.web.cern.ch/UHH2integration/test1171/mc_2017v2_TTSemiLeptonic_figs/jetsAk4CHS.btag_DeepFlavour_uds.pdf
Was the pre-computed one valid in the first place?

- how come DeepCSV does not agree for `jetsAk4Puppi` in anything other than 2018: e.g. 
https://uhh2-integration.web.cern.ch/UHH2integration/test1171/mc_2017v2_TTSemiLeptonic_figs/jetsAk4Puppi.btag_DeepCSV.pdf
vs
https://uhh2-integration.web.cern.ch/UHH2integration/test1171/mc_2018_TTLeptonic_figs/jetsAk4Puppi.btag_DeepCSV.pdf

(I realise DeepCSV isn't valid for PUPPI, but it still seems odd that it isn't even reproducible - I'd like to check there is nothing wrong with how we do PUPPI jets)

raggleton:
"Solved" by reverting to using values in Miniaod, see #1185

PR #1173, opened 2019-02-27
Lumi & PU files

anmalara:
This PR is to add the new files fulfilling the official recommendations for 2018. 2017 should be the same but I recalculate everything with the new CMSSW release, therefore there could be some slight difference. Btw, those should be the latest one.
Corrected also the preprocessor directives for JEC.


common/src/JetCorrectionSets.cxx
```diff
@@ -10,12 +10,9 @@ More details in UHH2/common/include/JetCorrectionsSets.h
 #define SET_NEWSTRING_DATA(tag,ver,jetCollection,sample,runName,Correction)                                                                                   \
 std::string("JECDatabase/textFiles/")+#tag+#runName+"_V"+#ver+"_"+#sample+"/"+#tag+#runName+"_V"+#ver+"_"+#sample+"_"+#Correction+"_"+#jetCollection+".txt",  \
 
-#define SET_NEWSTRING_DATA_2018(tag,ver,jetCollection,sample,runName,Correction)                                                                                   \
+#define SET_NEWSTRING_DATA_2018(tag,ver,jetCollection,sample,runName,Correction)                                                                                            \
 std::string("JECDatabase/textFiles/")+#tag+"_Run"+#runName+"_V"+#ver+"_"+#sample+"/"+#tag+"_Run"+#runName+"_V"+#ver+"_"+#sample+"_"+#Correction+"_"+#jetCollection+".txt",  \
 
-#define SET_NEWSTRING_DATA(tag,ver,jetCollection,sample,runName,Correction)                                                                                   \
```
raggleton: Is this not still needed?

anmalara:
It’s a copy of the first line, isn’t it?
I was unable to compile with that actually.

PR #1180, opened 2019-03-06
[102X] Add JEC V7 for JERC studies with 2018 data

karavdin:
[only compile]


raggleton:
I think this got killed because of the RAM usage - for some reason our compilation eats up RAM (like easily 8GB or even more). (This is probably related to why it takes so long as well)
I don't know if the runner machines have some limit before killing the job? I've now added in a call to `ulimit` and `time -v` to hopefully track a bit better the machine limits & usage limits. Will try the job again, let's see what it's doing.

raggleton:
I have a theory that all the `SET_JECFILES_*` in `JetCorrectionSets.cxx` are ultimately resolved into horrible string concatenations that are the root of all our woes.

If you run the preprocessor, expanding them (`g++ -E ../src/JetCorrectionSets.cxx -I/nfs/dust/cms/user/aggleton/UHH_Dev/102X_4/CMSSW_10_2_10/src/`) you can see lots of:

```
const std::vector<std::string> JERFiles::Summer16_07Aug2017_V11_L123_AK4PFchs_MC = { std::string("JECDatabase/textFiles/")+"Summer16_07Aug2017"+"_V"+"11"+"_"+"MC"+"/"+"Summer16_07Aug2017"+"_V"+"11"+"_"+"MC"+"_"+"L1FastJet"+"_"+"AK4PFchs"+".txt", 
std::string("JECDatabase/textFiles/")+"Summer16_07Aug2017"+"_V"+"11"+"_"+"MC"+"/"+"Summer16_07Aug2017"+"_V"+"11"+"_"+"MC"+"_"+"L2Relative"+"_"+"AK4PFchs"+".txt", 
std::string("JECDatabase/textFiles/")+"Summer16_07Aug2017"+"_V"+"11"+"_"+"MC"+"/"+"Summer16_07Aug2017"+"_V"+"11"+"_"+"MC"+"_"+"L3Absolute"+"_"+"AK4PFchs"+".txt", };
```

There are about 400 of these, each of which does 16 `+`, for 3 or more strings. So altogether it's doing at least 20K concats. string concat like this used to be (and maybe still is?) not so efficient - it is probably expanding memory each time it sees a `+`, hence ballooning up.

To test this, I replaced all the `SET_JECFILES_*` with all the actual vectors of strings, each already concatenated into a single string. e.g.
```
const std::vector<std::string> JERFiles::Autumn18_V7_L123_AK4PFchs_MC = { 
"JECDatabase/textFiles/Autumn18_V7_MC/Autumn18_V7_MC_L1FastJet_AK4PFchs.txt", 
"JECDatabase/textFiles/Autumn18_V7_MC/Autumn18_V7_MC_L2Relative_AK4PFchs.txt", 
"JECDatabase/textFiles/Autumn18_V7_MC/Autumn18_V7_MC_L3Absolute_AK4PFchs.txt", };
```

Testing with `/usr/bin/time -v make -j12`:

Before:
```
	User time (seconds): 1532.48
	Elapsed (wall clock) time (h:mm:ss or m:ss): 8:26.21
	Maximum resident set size (kbytes): 13207836
```

That's 13GB of RAM being used during compilation 😱 

After:
```
	User time (seconds): 855.03
	Elapsed (wall clock) time (h:mm:ss or m:ss): 2:23.06
	Maximum resident set size (kbytes): 1350728
```
Now down to 1.3 GB of RAM 😀 

So there's a massive improvement in memory & compilation time. So now we need a better way to do this string concatenation. Will handle in separate PR.

PR #1184, opened 2019-03-07
Module to run a Neural Network

anmalara:


common/src/NeuralNetworkBase.cxx
```diff
@@ -0,0 +1,327 @@
+#include <iostream>
+#include <fstream>
+#include <string>
+#include <sstream>
+
+#include "UHH2/HiggsToWWTagger/include/NeuralNetworkBase.hpp"
```
raggleton: Need to replace the `HiggsToWWTagger` with `common`?

common/include/NeuralNetworkBase.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+  Matrix2D MatrixMultiplication( Matrix2D A, Matrix2D B);
+  Matrix2D MatrixAddition( Matrix2D A, Matrix2D B);
+  Matrix2D ApplyActivation( Matrix2D A, TString activation);
+  Matrix2D getOutputs() {return outputs;};
+
+protected:
+
+  TString path;
+  TString fname = path+"mymodel.txt";
```
raggleton: Might be good idea to have a "/" preceeding each filename? e.g . `path+"/mymodel.txt"` so that the user doesn't have to worry about a missing trailing `/` in `path`? I guess one would then also need to check if `path == ""` then set `path="."` (or similar).

raggleton:
I've just found out about the lwtnn library: https://github.com/lwtnn/lwtnn which looks like it takes in keras JSON and can then apply it in C++ code. The track DNN people use it: https://github.com/cms-sw/cmssw/pull/21717/files


May be easier than doing matrix multiplication yourself?

karavdin:
According to log [1] issue is missing DeepCSV vars. 
We don't store probb and probbb separately for DeepCSV:
https://github.com/UHH2/UHH2/blob/RunII_102X_v1/core/include/Jet.h#L90
Is it required for this module?

[1] https://gitlab.cern.ch/raggleto/UHH2-integration/-/jobs/3540155

karavdin:
> I've just found out about the lwtnn library: https://github.com/lwtnn/lwtnn which looks like it takes in keras JSON and can then apply it in C++ code. The track DNN people use it: https://github.com/cms-sw/cmssw/pull/21717/files
> 
> May be easier than doing matrix multiplication yourself?

This looks interesting too. We should keep it in mind! 
Our own implementation is not bad as soon as it works. And it seems it does. Maybe @anmalara can add couple of plots illustrating validation to this PR?

anmalara:
> According to log [1] issue is missing DeepCSV vars.
> We don't store probb and probbb separately for DeepCSV:
> https://github.com/UHH2/UHH2/blob/RunII_102X_v1/core/include/Jet.h#L90
> Is it required for this module?
> 
> [1] https://gitlab.cern.ch/raggleto/UHH2-integration/-/jobs/3540155

It was store in 94X. This module shouldn't be pushed in the 102X branch yet.

> Our own implementation is not bad as soon as it works. And it seems it does. Maybe @anmalara can add couple of plots illustrating validation to this PR?

Which plot do you have in mind? Comparing NN in python and C++ implementations is not trivial. I did a manual check for a couple of events.

karavdin:
@anmalara, didn't notice the PR is for 94X_v3 branch, but change in Jet.h is done here too:
https://github.com/UHH2/UHH2/blob/RunII_94X_v3/core/include/Jet.h#L71

As a comparison plot something like output node response from tensorflow vs response from this module would be good.

anmalara:
> @anmalara <https://github.com/anmalara>, didn't notice the PR is for 94X_v3 branch, but change in Jet.h is done here too:
> https://github.com/UHH2/UHH2/blob/RunII_94X_v3/core/include/Jet.h#L71 <https://github.com/UHH2/UHH2/blob/RunII_94X_v3/core/include/Jet.h#L71>As a comparison plot something like output node response from tensorflow vs response from this module would be good.

Good to know that the class has changed also in that branch.
Btw, for the time being I will not use that version of the files in 102X. It’s too much user dependent.
The commit was only to make Arne able to run the code quickly. But the module need more work before it can be largely used.
I will make new files later.

PR #1163, opened 2019-02-21
[102X] add b-taggers WPs for AK4 jets, rename Jet::btag_DeepFlavour() to Jet::btag_DeepJet(), specify era of process 

karavdin:
add working points for b-taggers for 2016, 2017 and 2018;
change list of discriminators to enable storage of DeepCSV for AK4PUPPI in 2016v2
rename `Jet::btag_DeepFlavour()` to `Jet::btag_DeepJet()`

Check with Z'->ttbar selection for tight WP of each b-tagger on AK4PUPPI for samples of different years:
b_taggers_zprime

PR #1185, opened 2019-03-07
[102X] copy most of slimmed b-taggers from miniAOD

karavdin:
update pfDeepFlavourJetTags for all years as before and the rest only for 2016v2. For other years it should be copied from miniAOD. This should "fix" issue with changed DeepCSV for 2017 discussed in https://github.com/UHH2/UHH2/pull/1163

Note for the future: BTV POG does not proved support for b-taggers in PUPPI jets yet, therefore we store discriminators calculated in miniAOD for PUPPI jets if it's possible (i.e 2017 and 2018) and add it only for 2016v2, where it isn't calculated yet. However one should be careful with using it in 2016, since it isn't clear if the DNN training is applicable to 2016v2 data.

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -1188,9 +1188,12 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
         if useData:
             jetcorr_list.append("L2L3Residual")
 
-        discriminators = bTagDiscriminators[:]
-        discriminators.extend(ak4btagDiscriminators)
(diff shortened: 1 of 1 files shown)


raggleton:
This change is because of complications trying to recalculate the DeepCSV, combinedSecondaryVertex, combinedSecondaryVertexMVA variables on PUPPI jets. Since they aren't actually supported on PUPPI jets by the BTAG POG (only CHS are), we now just copy whatever is in MiniAOD for those. The only special case is 2016v2, for which we do recalculate those variables.

PR #1193, opened 2019-03-13
[102X] Puppi weights + tidying for jet constituents

raggleton:
Store PUPPI weights for jet constituents. Also tidied up the duplicate methods used for (gen)jet constituent storage, now only one copy in NtupleWriterJets.cxx.

Also improved on constituent check - previous checked deltaR and deltapT == 0f, which often fails for floats. Now checks relative difference is < 1E-6 OR absolute difference is < 1E-6, whichever is larger. The latter absolute diff is necessary for deltaR == 0 comparison, since both the absolute and relative differences are ~1E-7 or so, but the relative diff condition must be < 1E-13, which is never satisfied.

This should hopefully shave off some duplicates when storing constituents.

### DIFF
--- a/core/include/Utils.h
+++ b/core/include/Utils.h
@@ -85,4 +85,7 @@ private:
     std::vector<std::vector<std::string> > rows;
 };
 
+/// Compare 2 floats - better than == does to floating-point issues
+bool closeFloat(float a, float b, float maxRelDiff=1E-6);
+
 }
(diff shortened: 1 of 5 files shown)


raggleton:
So can see now a reduction in the # gen particles & pf particles stored when storing jet constituents due to now successful matching. jetsAk4CHS constituents don't change, which is good. Nice reduction in filesize as well.

PR #1195, opened 2019-03-13
[102X] Get PUPPI userFloats by matching strings to ignore producer's name

raggleton:
Better way to get the PUPPI multiplicities from userFloats.

Also now does it for the subjets of puppi topjets.

To save time looping through all userFloat names unnecessarily, we now also do this only for jet collections with "puppi" in their name (case independent). This removes the need for the PUPPI specific producer to have the same name as the jet collection (my previous horrible hack that caused much woe, I don't know why I didn't think of this earlier).

### DIFF
--- a/core/plugins/NtupleWriterJets.cxx
+++ b/core/plugins/NtupleWriterJets.cxx
@@ -32,6 +32,8 @@
 #include "DataFormats/BTauReco/interface/BoostedDoubleSVTagInfo.h"
 #include "DataFormats/Math/interface/deltaR.h"
 
+#include <algorithm>
+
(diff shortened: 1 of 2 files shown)


raggleton:
Difference in plots are (a) for CHS jets, where the puppi defaults are now those in Jet class (0s), rather than in the fill_jet_info (-1s); (b) for PUPPI subjets, where it is now using the default in fill_jet_info than in Jet class. 

Realise we purposely don't run the PUPPISpecificProducer for subjets as it wasn't needed/too tricky to configure (although now we don't care about the producer name, maybe easier?)

PR #1196, opened 2019-03-14
[102X] cleanup HOTVR headers from XCone producers

raggleton:
Dunno why they were in there, not necessary. Also had to add in std namespace because the HOTVR.hh had it in (see https://github.com/UHH2/HOTVRContrib/pull/2)

[only compile]


raggleton:
Failure was in the reference build, since the HOTVR plugin code changed, so it's expected to fail. The PR build worked fine, so will merge this

PR #1199, opened 2019-03-18
[102X] Rekey PUPPI jet constituents

raggleton:
This "rekeys" the constituents of PUPPI jets to replace them with their equivalent from the packedPFCandidates (this is what is done for slimmedJetsPuppi in MiniAOD). This way, when we come to store the jet constituents, we do not end up storing "duplicate" PUPPI & PF particles that differ only in whether they have the PUPPI weight applied. Since we now also store PUPPI weight (#1190) the user can still do things with the jet constituents, so long as they remember to use the PUPPI weight.

Have done for jetsAk8Puppi, jetsAk8PuppiSubstructure_SoftDropPuppi, xconePuppi, hotvrPuppi, and all the xconeDijetPuppi collections. 

- Pros: now uniform behaviour across all PUPPI jet collections, save space by not duplicating particles.
- Cons: now people have to choose to apply PUPPI weight or not.

Technical note: for the TopJets, I have only rekeyed the subjets and the ungroomed fatjet (and actually I rekey CHS as well inside `add_fatjet_subjets()`). This is because these are the only jet constituent we care about. Additionally, I have left re-keying as late as possible so that e.g. b-tagging calculations are not affected.

NB in MINIAOD production this is done by either `PATJetSlimmer` or `JetSubstructurePacker`. These require a map to PF particles, which we don't have, so I had to write a small producer to handle this instead. On the plus side, it actually allows rekeying from PUPPI to PF, or vice-versa, incase we ever want to convert the slimmedJetsPuppi constituents to PUPPI particles.

NB2 also updated ExampleJetConstituents to add in flags to apply PUPPI weights or not.

### DIFF
--- a/core/plugins/RekeyJets.cc
+++ b/core/plugins/RekeyJets.cc
@@ -1,0 +1,137 @@
+// -*- C++ -*-
+//
+// Package:    UHH2/RekeyJets
+// Class:      RekeyJets
+//
+/**\class RekeyJets RekeyJets.cc UHH2/RekeyJets/plugins/RekeyJets.cc
+
+ Description: Replace duaghters of jet with those matched by key from another collection
+  e.g. replace puppi jet daughters with equivalent particles from packedPFCandidates
+
+*/
+//
+// Original Author:  Robin Aggleton
+//         Created:  Tue, 23 Jan 2018 13:27:50 GMT
+//
+//
+
+
+// system include files
+#include <memory>
+
+// user include files
+#include "FWCore/Framework/interface/Frameworkfwd.h"
+#include "FWCore/Framework/interface/stream/EDProducer.h"
+
+#include "FWCore/Framework/interface/Event.h"
+#include "FWCore/Framework/interface/MakerMacros.h"
+
+#include "FWCore/ParameterSet/interface/ParameterSet.h"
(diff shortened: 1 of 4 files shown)


karavdin:
Wonderful! One can clearly see change in PF candidate indexs for PUPPI jets for mc_2018_TTHadronicAllConstits and less PF candidates stored. It's sad that the reduction in size is not so significant, which is somehow also expected since number of duplicates was not that high.

raggleton:
Hmm yes I was hoping for a slightly larger reduction in size - the PFPArticles have only dropped by 3% in size. Will have to double check if there is some failure in matching or rekeying.

I also realise I was a bit disingenuous with my description above: this now does the rekeying _before_ we do deepflavour & deepboosted tagging, and you can see that this does slightly change those discriminants. However it is still setup correctly, in that it knows to apply the PUPPI weights in the deep boosted modules (like for slimmedJets*). I guess one could do it afterwards to be ultra correct, although I don't know if those discriminants use subjet information, which is already rekeyed at that point.

raggleton:
Ack, also discovered that the subjets of XCONE & HOTVR TopJets aren't being rekeyed properly - they still store puppi-weighted constituents.

PR #1202, opened 2019-03-20
[102X] Add rekey options to HOTVR & XCone producers, use it on PUPPI jets

raggleton:
Instead of using RekeyJets producer + messy hacks, move rekeying for HOTVR & XCone into their respective producers. The now actually rekeys their subjets as well, which was missing from #1199 .
Rekeying subjets is easiest for these collections done in their producers, due to the fact that a pat::Jet does not allow you to remove subjets once added, so you end up with a messy set of subjets.

With this, all PUPPI jet collections (& their subjets) should be correctly rekeyed against packedPFCandidates.

### DIFF
--- a/core/plugins/HOTVRProducer.cc
+++ b/core/plugins/HOTVRProducer.cc
@@ -63,8 +63,9 @@ class HOTVRProducer : public edm::stream::EDProducer<> {
     ~HOTVRProducer();
 
     static void fillDescriptions(edm::ConfigurationDescriptions& descriptions);
-  // This method copies the constituents from the fjConstituents method to an output of CandidatePtr's. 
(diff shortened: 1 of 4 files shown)


raggleton:
Looks better now. Can confirm that recalculating subjets with PUPPI weights gives good agreement with original subjet.

BTW size/event total drop is from 37.822 to 36.867, so about 2.5%. I think that alot of the particles are already clustered & saved in the ak4 CHS & PUPPI jets with their low thresholds, so perhaps we can't expect such a large saving.

raggleton:
Not quite sure why the pfcand_indexs changed for the main fatjet though, it should be doing exactly the same thing. Maybe saving the subjets slightly re-jiggled the constituent order in event.pfparticles (woudl also explain why xconeCHS is affected, despite not changing in the config!)

PR #1204, opened 2019-03-21
[102X] Simplify genparticle pruning into 1 module to possibly save some RSS memory

raggleton:
So I _think_ compactifying all our gen particle pruning into 1 module instead of 2 may save us some memory. The content should be identical. I don't think it saves any processing time unfortunately.

All tests were done with TTHadronic 2018 MC with the "allConstits" config. I did it twice for the default config, and twice for this PR.  The two "default" runs are on different machines at different times, the compact pruning lines were done one after the other.

RSS does show some differences: here is the RSS **relative** to the 1st recorded event:

image

Note that it doesn't even agree in repeated running (e.g. blue & orange, red&green)...a very tricky metric 😕 

Looking at absolute RSS:

image

Not quite sure what to take away from the absolute values - running the same default config, the RSS drop by 100MB (blue  & orange) for no obvious reason?!

VSIZE doesn't change, although I think RSS is the real limitation for CRAB jobs:

image

I think overall though it _might_ help save anywhere from 50 to 100 MB ? It doesn't look like it does any harm anyway. 🤷‍♂️ Not quite the magic fix we need unfortunately

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -302,11 +302,11 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
     # GEN PARTICLES
     #
     # The 13TeV samples mainly use Pythia8 for showering, which stores information in another way compared to Pythia6; in particular,
-    # many intermediate particles are stored such as top quarks or W bosons, which are not required for the analyses and makle the code more complicated.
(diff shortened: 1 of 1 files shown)


raggleton:
No changes to genparticle content. I did think about adding in memory metric to this automated testing, but you'd need at least ~2.5K events to actually show differences, which would increase testing time massively. Plus, as shown above, it can change from run to run, so like the timing, not the most reliable metric.

PR #1205, opened 2019-03-21
add Selection for recalculated metfilter related to ecalBadCalib

anmalara:
[only compile]

### DIFF
--- a/common/include/NSelections.h
+++ b/common/include/NSelections.h
@@ -12,7 +12,7 @@
 //
 // The modules allow two modes of operation:
 //  * no id criterion is applied, so all currently available objects are counted.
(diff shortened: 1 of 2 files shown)


raggleton:
We don't need to run the full chain of tests for this since it only affects post-ntuple processing. Can you add `[only compile]` to your original PR text please? Will then stop & retrigger tests

raggleton:
The other MET filters will need updating (e.g. removing `Flag_BadChargedCandidateFilter` for all years), but that can happen in a separate PR

anmalara:
> The other MET filters will need updating (e.g. removing `Flag_BadChargedCandidateFilter` for all years), but that can happen in a separate PR

Agreed. My guess is that that filter is not ready yet, but it will be soon. Let's check later on.

PR #1209, opened 2019-03-27
[102X] Use unique_ptr to avoid raw ptr leaking in GenXConeProducer

raggleton:
This solves a long-standing memory (RSS) leak that plagued MC samples. Short (2K evt) tests look promising, will run much long ones as well to check.

### DIFF
--- a/core/plugins/GenXConeProducer.cc
+++ b/core/plugins/GenXConeProducer.cc
@@ -310,12 +310,12 @@ GenXConeProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup)
 
     // Run second clustering step (subjets) for each fat jet
     vector<PseudoJet> subjets;
-    //ClusterSequence * clust_seq_sub;
+    std::unique_ptr<ClusterSequence> clust_seq_sub;
     if (enoughParticles && doSubjets) {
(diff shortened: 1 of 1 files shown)


raggleton:
Running over 5K ttbar 2018 MC events: (this PR is the light blue "ttbar MC, all (constit, ptr fix)" line.
You can see that before in e.g. orange, red, light green, there was a general growth of RSS with # events. Whereas with this fix, it is now more constant (modulo the occasional large jumps).
There are also a variety of other lines there switching bits on/off, plus a data sample for comparison.

Ignore the fact that the total RSS is larger - at some point it just jumped up overall, but more important is that it doesn't grow constantly vs # events.

image
image

(Technical history: memory data taken using the `SimpleMemoryCheck` service. I discovered this by seeing that data wasn't growing, so something gen-related. Tried turning off geninfo, genjets, gentopjets. Narrowed it down to GenHOTVR and/or GenXCone. Turned off actually adding those objects to output TTree, but still got Handle to ensure modules ran. Then enabled each one individually to check which one (or both) was leaking. Then looked for immediate suspect in modules, `new`. Found this one, fixed it, reran test, bingo 🎉 )

PR #1211, opened 2019-03-28
[102X] Update default crab cfg, unitsPerJob

raggleton:
[ci skip]

NB jobs run about 4K evt/hour, so 16K is ~ 4hours runtime.


raggleton:
Following QCD tests which showed mem usage at ~ 2GB (with everything enabled), happy to reduce maxMemory to 2500 MB. Also increase unitsPerJob to 24000 (~6 hrs) so that there are fewer jobs in total to make things a bit more efficient & avoid getting in long queues.

PR #1214, opened 2019-04-01
Added at csh version of the script

drberry85:
Hey Robin and Anastasia,

I added a csh version of the installation script.

Doug

### DIFF
--- a/scripts/install.csh
+++ b/scripts/install.csh
@@ -33,7 +33,7 @@ autoreconf -f -i  # needed to avoid 'aclocal-1.15' is missing on your system
 make $MAKEFLAGS
 # make check  # fails for siscone
 make install
-cd ..
+cd ../..
 
 # Add fastjet-config to PATH
 setenv PATH "${FJINSTALLDIR}/bin":${PATH}
--- a/scripts/install.csh
+++ b/scripts/install.csh
@@ -1,3 +1,5 @@
+#!/usr/bin/env tcsh -e
+
 # Create default make args for parallel jobs
 set maxcores=12
 set cores=`grep processor /proc/cpuinfo | wc -l`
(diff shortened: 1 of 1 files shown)


raggleton:
Hi, thanks Doug. For automated testing purposes I can add a test for the csh install script as well to check it deploys OK. Silly question, but how do you exactly run it? `csh install.csh`?

BTW should there be a shebang at the start of this script?

drberry85:
Hey Robin,

This is a non-executable version of the script. I'll upload an executable
version with a shebang.

Doug

drberry85:
Hey Robin,

I've never done an automated build test on CERN git labs before. Would you recommend a tutorial or example I can follow?

Thanks,
Doug

raggleton:
https://docs.gitlab.com/ee/ci/README.html is the general Gitlab CI documentation. I was just going to add an additional part to our existing setup to also run your new `install.csh` script, and then check it compiles fine (like you were checking out a fresh new copy). 

I'm not familiar with tcsh - how do you run this script when you want to setup a release? We `source` our bash script, since it then sets up all the `CMSSW_VERSION`, `SCRAM_ARCH` etc in the current shell, but I don't know what the equivalent is for tcsh?

drberry85:
Robin,

You run it the same way as ```install.sh````. You just source ```install.csh``` in a tcsh (or csh) environment. That's why I didn't make the script executable or add a shebash, because all you need to do is source it.

Doug

raggleton:
Ok, in that case I would need to setup a test job that first sets up a csh shell, then convert the commands & scripts that do pre/post install to csh. At the moment this isn't a priority, so I'll merge this, and try and come back to it at some point.

drberry85:
Hey Robin,

I'm not sure hot to setup a test job in csh/tcsh. If you should me your bash scripts, I can translate them to tcsh.

Doug

PR #1216, opened 2019-04-03
[102X] Add CRAB output subdirectory using year from pset filename

raggleton:
[ci skip]

e.g. now stores in `/store/user/xxxx/RunII_102X_v1/2018/` for `ntuplewriter_mc_2018.py`. Should handle `2017v2` as well OK.


raggleton:
Currently testing this with some CRAB jobs just to check all runs OK, then will merge

PR #1235, opened 2019-04-11
add PU distribution for MC

anmalara:


raggleton:
What cross-section (or even better, the actual command) did you use for the pileupCalc.py so that we have it for reference?

BTW in future, for PRs that don't need compilation or testing like this, you should add [ci skip] to your PR message.

raggleton:
Sorry, didn't mean pileupCalc.py, meant what reference did you use for the MC distribution? a CMSSW py config, or from a MC sample?

anmalara:
This is calculated using the CMSSW config file.
I’ve checked also the alternative approach for QCD sample and those are basically identical.

PR #1236, opened 2019-04-11
Inclusive Single Top Sample

drberry85:
Robin and Anastasia,

Here are the inclusive tW single top samples. Still waiting on the no hadronic.

Doug


raggleton:
BTW for PRs like this, that don't affect the compilation of the code nor the ntuples, you can add `[ci skip]` to your PR message to avoid starting the automated testing

PR #1246, opened 2019-04-22
[102X] Add ability to store extra trigger bits; use for BadPFMuonFilter instead of in veto mode

raggleton:
- Adds `extra_trigger_bits` argument to NtupleWriter to allow storing of arbitrary bits (any module that produces a `bool`). These get stored like other trigger bits, but as `Extra_*` to avoid clashing with existing trigger names
- Use this to store BadPFMuonFilter now, which is now run as a tagger instead of filtering events
- Update `CommonModules` to use this new flag for 2016v2 samples


raggleton:
Should note that this solution means that ntuples are still backwards-compatible (the `Event` class hasn't changed structure), and that this only affects the 2016v2 configs.

Note to self: in theory one could also store the ECAL bad calib bit in this way, instead of as a separate bool. But since ntuples have already been made using the separate bool, keep it that way to avoid screwing things up.

PR #1248, opened 2019-04-24
 [ci skip] add puppi correction file for soft drop mass calculator

IreneZoi:
The same file has been added for 2016, 2017 and 2018. Need to check for eventual updates
[ci skip]


raggleton:
You need to update your target branch to `RunII_102X_v1`. Please also add `[ci skip]` in your main PR message, since we don't need to compile or test anything.

raggleton:
For future reference: although the `puppiCorr.root` file has been to the separate 2016/7/8 directories, they're all the same 2016 file since they don't exist yet for 2017, 8. At the moment they're the best guess for 17,18. Will update those when new versions become available (no ETA).

PR #1250, opened 2019-04-25
Enriched 2018 QCD MC XML [skip ci] 

drberry85:
Robin and Anastasia,

XML files for QCD Mu and EM enriched. It looks like there are no bcToE samples in 2018 MC. [skip ci]

Doug


karavdin:
Hi Doug,
Thanks for the files. I must mention the command to skip ci test wrong, it should be [ci skip].

Cheers,
Anastasia

PR #1251, opened 2019-04-25
[skip ci]

titasroy89:


karavdin:
Hi Titas,
Thanks for the files! Most probably I mentioned the command to skip ci test wrong today. 
It should be [ci skip].

Cheers,
Anastasia

PR #1262, opened 2019-05-03
[102X] Remove duplicate XMLs

raggleton:
[ci skip]

These were added to `common/datasets/RunII_102X_v1/2018` in #1257, but were also added to `scripts/crab`. I am assuming they're safe to delete from `scripts/crab` @karavdin @titasroy89 ? The exact same XMLs are in `common/datasets/RunII_102X_v1/2018`


titasroy89:
Yes, these are duplicates from my area. Please feel free to remove it. I will clean it up in my next commit as well. thanks!

PR #1263, opened 2019-05-03
fixed softdropmass() for GenTopJet returning always zero

jngadiub:


raggleton:
Thanks, looks like from e.g. https://uhh2-integration.web.cern.ch/UHH2integration/test1263/mc_2016v2_TTbar_figs/genjetsAk8SubstructureSoftDrop.softdropmass.pdf it fixes things (I was silly before - didn't check the plots 😞 )

PR #1275, opened 2019-05-08
[102X] year & run auto-switchers for analysis modules

raggleton:
[only compile]

Add 2 new general classes: `YearSwitcher` and `RunSwitcher`. Each is designed to encapsulate away the logic of switching modules based on year and run #. The idea is the user sets up all the relevant modules at the start in their analysis constructor, then in their analysis `process()` they just call the `process()` method of these classes and don't have to worry any more. An example of usage of each of these, and using them in combination, is included in `ExampleModuleYearRunSwitcher`.

This way the same analysis can be used for different years, and the modules can easily be set up - each accepts anything inheriting from `uhh2::AnalysisModule`.

`YearSwitcher` can be setup for any 2016 dataset, or specialise for 2016v2, v3 separately. If the user does `setup2016()` then `setup2016v3()`, then a 2016v3 dataset will use the latter, whilst a 2016v2 dataset will use the former. Similarly for 2017, v1 & v2.

A `RunSwitcher` instance can be setup for a given year. Here it will ignore `v*`, so e.g. treats `2016v3` as `2016`. Then the user can add modules for given run periods, specifying the run period with a string to allow for looping (see example module L99, 104, 109)

C++ technicality: I'm not entirely sure using a `shared_ptr` is the correct way to aggregate here - I am trying to cover the scenario when the user creates the object in the `setup*` method calls, but also when they have already created the object, and this class needs it to stay alive.

Feedback on the design is welcome!

A target for these classes is the JECs...

### DIFF
--- a/common/include/Utils.h
+++ b/common/include/Utils.h
@@ -119,6 +119,15 @@ Year extract_year(const uhh2::Context & ctx);
 
 
 /**
+ * Run period names for each year
+ */
+const std::vector<std::string> runPeriods2016 = {"B", "C", "D", "E", "F", "G", "H"};
+
+const std::vector<std::string> runPeriods2017 = {"B", "C", "D", "E", "F"};
+
+const std::vector<std::string> runPeriods2018 = {"A", "B", "C", "D"};
+
+/**
  * Map run periods to run numbers for each year
  *
  * All pairs for a run period are inclusive of both lower & upper numbers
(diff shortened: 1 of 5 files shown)


raggleton:
Don't merge this yet - need to fix dodgy handling of raw pointers with shared_ptr

raggleton:
OK am happy with how pointers are handled now - we don't use any raw pointers now, so no chance of accidentally deleting them (or not). The slight cost is that the user now has to use e.g. `make_shared<T>()` instead of just `new T()`, or pass in a `shared_ptr`. Hopefully the example module + the upcoming changes to CommonModules will be a good reference for users.

PR #1287, opened 2019-05-15
[102X] Use Year/RunSwitcher in CommonModules for JECs, add 16/18 JECs

raggleton:
This updates CommonModules to use the new YearSwitcher & RunSwitcher classes added in #1275. This means that it can now apply 2016 & 2018 JECs to `event.jets` for those year datasets.

(Although I realise it's hard-coded to AK4PFCHS - something to fix in another PR to handle PUPPI)
 
[only compile]


raggleton:
Also I used the latest JEC sets for 16 & 18 as specified in JetCorrectionSets.h, but perhaps they need updating?

PR #1305, opened 2019-05-28
[102X] Move to LZMA9 compression

raggleton:
Change default Ntuple compression from default ZLIB1 to most compressed LZMA9. This saves us ~25-30% in filesize, with a small penalty in CPU/RAM usage.
Needed as we're rapidly running out of disk space, so this should be used for all subsequent productions.

More detailed analysis: [20190528_UHH2_compression.pdf](https://github.com/UHH2/UHH2/files/3226833/20190528_UHH2_compression.pdf)

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -2095,8 +2095,8 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
                                     #    TestKey = cms.string("TestValue")
                                     #),
                                     fileName=cms.string("Ntuple.root"),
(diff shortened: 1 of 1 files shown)


raggleton:
The gitlab machines have moved to SL7, which is in conflict with our setup for SL6, failing the jobs. This might not be an easy fix, as I have to now figure out how to find a docker image with SL6...

raggleton:
Great, can see ~20 - 25% filesize reduction to show it is indeed working (% here is a bit smaller than in other study as we only have 500 events).

PR #1306, opened 2019-05-28
[102X] Muon scale factors for 2018

froehlia:
[only compile]

* Added .root files containing muon scale factors ([see here](https://twiki.cern.ch/twiki/bin/view/CMS/MuonReferenceEffs2018#Scale_Factors_with_statistical_e))
* Added a line in the muon scale factors module to allow to pass the histogram name directly


raggleton:
OK fixed the SLC6 image problem, now compiles fine (turned off github messages for this test)

PR #1324, opened 2019-06-06
Simplified macro for calculate root entries. Add ntuplewriter macro

anmalara:
Add ntuplewriter macro for 2016 v2 with leadingjetConstits

[only compile]


karavdin:
@anmalara, could you remove "readaMCatNloEntries.py" from this PR, please? As far as I can see it just reverts back to previous version of readaMCatNloEntries.py script. If there is anything useful I missed, please "merge" the changes locally and commit updated file. Thanks!

anmalara:
the update is commenting empty files while calculating the weights. Moreover, in case one updated the aforementioned files and re-run the script, those should be put back to the list.

PR #1331, opened 2019-06-18
[102X] added JetID for removing Jets according to EtaPhi Map

stalbrec:
Hi @karavdin! Here is the EtaPhi Cleaning JetId i mentioned. Is this something, that is useful?
[only compile]


raggleton:
It might be easier to find the bin using the builtin methods:

```
TAxis *xaxis = h_map->GetXaxis();
TAxis *yaxis = h_map->GetYaxis();
Int_t binx = xaxis->FindBin(jet.eta());
Int_t biny = yaxis->FindBin(jet.phi());
```

This would also be safer, since I imagine the binning could be non-constant width (especially if mapping ECAL/HCAL cells)?

raggleton:
Do we also want a constructor that instead takes a filename, and then automatically opens it & `Get()`s a TH2? Perhaps easier for the user than having to deal with opening file, checking TH2 exists, etc.

stalbrec:
I actually went with passing the filename instead initially, but ran into trouble because it opened a file for each jet and then the code crashed.

karavdin:
Hi Steffan,
This indeed would be very useful. Not sure why it didn't work with passing only file name string. Did you try it in similar way as in MCWeight ? E.g in the class below:
https://github.com/UHH2/UHH2/blob/RunII_102X_v1/common/src/MCWeight.cxx#L538-L633 

The bins indeed might have flexible width, so I agree with Robin that FindBin is more safe approach.

@stalbrec, could you implement the changes discussed here, please? This is not very urgent, but will be useful for all people looking at jets beyond the tracker coverage.

raggleton:
So long as you put the opening of the TFile & getting of the TH2 into the `JetEtaPhiCleaningId` constructor like in `MCWeight`, and the user only constructs one copy of the object (e.g. in their analysis module constructor, like setting up the `JetCleaner` object here: https://github.com/UHH2/UHH2/blob/RunII_102X_v1/examples/src/ExampleModule.cxx#L68), then it should only open the file once when the class is constructed.

Also, can you add the ROOT file as well (if it isn't already in `common/data`), thanks

stalbrec:
Okay, thanks. I will change it according to your comments. 
Should I also add the name of the histogram as an optional parameter in the constructor?

raggleton:
Yeah that's probably a good idea (e.g. if someone wants to try using a different map file).

stalbrec:
I added the discussed changes. In UHH2 it compiles, but when i try to initialize a JetEtaPhiCleaningId in an analysis it does not compile and i get an error looking like [this](http://desy.de/~albrechs/comp.log).
In the analysis i did the follwoing:
In the Analysis header:
...
JetId EtaPhiId;
...
In the Constructor:
...
EtaPhiId= JetId(JetEtaPhiCleaningId("common/data/2018/hotjets-18runA.root"));
...

stalbrec:
I tried to use a normal pointer for the Histogram instead of a unique_ptr and now it works.
Would this be an option, or is it mandatory to use unique_ptr?

raggleton:
So the relevant bits of the error are:

```
error: use of deleted function 'JetEtaPhiCleaningId::JetEtaPhiCleaningId(const JetEtaPhiCleaningId&)'
```
and
```
note: 'JetEtaPhiCleaningId::JetEtaPhiCleaningId(const JetEtaPhiCleaningId&)' is implicitly deleted because the default definition would be ill-formed:
...
error: use of deleted function 'std::unique_ptr<_Tp, _Dp>::unique_ptr(const std::unique_ptr<_Tp, _Dp>&) [with _Tp = TH2; _Dp = std::default_delete<TH2>]'
```

The first error is that the code is trying to call the class' copy constructor. C++ creates a "default" copy constructor for a class if you don't explicitly specify one, and its behaviour is to copy each member variable. However, as we are using a unique_ptr here (which doesn't support copying, as that's the nature of a single-ownership pointer, that's the `use of deleted function 'std::unique_ptr<_Tp, _Dp>::unique_ptr(const std::unique_ptr<_Tp, _Dp>&) ` part), it automatically deletes the default copy constructor (hence the `note: ... is implicitly deleted...`)

This copying occurs in a few places. One is in the `clean_collection` signature, where the std::function is passed by value, not reference, hence why it gets copied: https://github.com/UHH2/UHH2/blob/RunII_102X_v1/common/include/Utils.h#L81

Another is in `JetCleaner` ctor: https://github.com/UHH2/UHH2/blob/RunII_102X_v1/common/src/CleaningModules.cxx#L11 where the member variable `jet_id` gets instantiated by copying the `jet_id_` argument (even though it is passed by reference).

In general, the way to resolve this is to use `std::move()` to move the resource, rather than copy it. 
However,  in this scenario, because we are really assigning it to a `std::function` (typedef of a `JetId`), `JetEtaPhiCleaningId` is really a functor, and there is a requirement in C++ that the **functor must have a copy constructor to be assignable to a std::function** (https://stackoverflow.com/questions/11768273/why-cant-c11-move-a-noncopyable-functor-to-a-stdfunction, https://en.cppreference.com/w/cpp/utility/functional/function). So if you really want to use unique_ptr here, you'd have to write your own copy constructor, which `Clone`s the TH2, e.g. `JetEtaPhiCleaningId(const JetEtaPhiCleaningId & o): h_map((TH2*) o.h_map->Clone()) {}`

In this situation, one easier option for now is just storing the object itself (rather than as any sort of pointer):

in JetIds.h:
```
class JetEtaPhiCleaningId{
 public:
    explicit JetEtaPhiCleaningId(const std::string & mapFilename, const std::string & mapHistname="h2hotfilter");
    bool operator()(const Jet&,const uhh2::Event&) const;
 private:
    TH2D h_map;
};
```

and in JetIds.cxx:

```
JetEtaPhiCleaningId::JetEtaPhiCleaningId(const std::string & mapFilename, const std::string & mapHistname)
{
  TFile map_file(locate_file(mapFilename).c_str());
  if (map_file.IsZombie()) {
    throw runtime_error("2D map file not found: " + mapFilename);
  }
  
  if (!map_file.GetListOfKeys()->Contains(mapHistname.c_str())) {
    throw runtime_error("2D map histogram not found in file");
  }
  h_map = *((TH2D*) map_file.Get(mapHistname.c_str()));
  h_map.SetDirectory(0);	
  map_file.Close();
}

```
then update the `operator()` accordingly.

(Note that I guessed it was a TH2D and not a TH2F...which ROOT may complain about if it's wrong, so please change as necessary.)

A raw pointer here is not great, because when the class gets deleted, it will only delete the pointer itself, but not the object to which it was pointing, thereby potentially leaking memory. We want our JetEtaPhiCleaningId class to own the memory used by `h_map` member variable, and be responsible for automatically clearing it up when it gets deleted.

Alternatively, one could also use try a shared_ptr, to enable the default class copy constructor.

(Sorry I realise that was long...I learnt a lot of new C++ things trying this out!)

stalbrec:
Thank you very much! I changed it as you suggested and also added the root files to `common/data`

PR #1337, opened 2019-06-21
[102X] Add == operator to Particle

raggleton:
Allow (in)equality testing between `Particle`s and all its derived classes (e.g. `Jet`, `Muon`, `GenParticle`, etc)

[only compile]

### DIFF
--- a/core/include/Particle.h
+++ b/core/include/Particle.h
@@ -5,7 +5,7 @@
 
 class Particle{
 public:
-     
+
    Particle(): m_charge(0), m_pt(0), m_eta(0), m_phi(0), m_energy(0){
    }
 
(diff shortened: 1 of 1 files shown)


raggleton:
Although I've just discovered that underneath it all, this uses the == operator of the LorentzVector, which boils down to a float equality of the pt/eta/phi/E components: https://root.cern.ch/doc/v610/GenVector_2PtEtaPhiE4D_8h_source.html#l00321 This relies on float equality, which can be problematic. 

I'm wondering if this need explicit handling by us... although I imagined that the use cases for the == in `Particle` is e.g. spotting duplicate particles in some list, rather than trying to match separately-created objects, e.g. a Particle created with pt=2.5 vs one created with pt=5./2.
Plus often people "match" based solely on deltaR, whereas this operator is more for identical object identification (i.e. one is a clone of the other).

Still, something to bear in mind

PR #1347, opened 2019-07-06
Add missing DiBoson samples for 2016

anmalara:
[ci skip]

With this PR, all DiBoson samples (inclusive and not) are produced for 2016 dataset.
They contain PF candidate too. The relative x-sections and n.events are already updated in the google doc.


raggleton:
I see that some of the samples, e.g. WWto4Q are Powheg but don't use weights in calculating the number of events - should it be used?

karavdin:
@raggleton, indeed Poweg samples contain weights not equal 1. 
@anmalara, could you re-calculate number of events with "False" argument for readaMCatNloEntries.py script, please?

anmalara:
That particular sample has all weight = 1. I have also re-calculate with the other procedure, and I got the same number. Similar conclusions hold for all the samples I pushed in the repo. But please let me know if you spot some error.

PR #1350, opened 2019-07-12
[106X] Changes for 106X Compatibility and changes in Ntuplewriter for UL 106X Samples

stalbrec:
In the UL17 106X MINIAOD samples the prescales and metfilters seem to be not (complety) stored.
So this version of the Ntuplewriter.cc stores the triggerResults but not the prescales and not the metfilterResults.

~Also in the ntuplewriter config i had to add:~

~`process.MyNtuple.year="2017v2"`~
~`process.MyNtuple.doL1seed=cms.bool(False)`~

~in order to get it working.~
^ These are now in ntuple_generator.py

Also: The TMVAEvaluator moved from CommonTools/Utils/interface/ (102X) to CommonTools/MVAUtils/interface (106X).

[ci skip]


raggleton:
FYI the whole automated testing procedure won't work for 10_6 currently - needs to change to use slc7 + some other fixes. For now, letting you guys check compilation etc. Would recommend using [ci skip] until it's been fixed

stalbrec:
Oh

stalbrec:
Yes, I forgot it would do it in 102. I also forgot to add my changes to install.sh. Will update asap
Also sorry for closing and reopening.

stalbrec:
I introduced the year `2018UL17` in ntuple_generator.py to get the correct "2018 setup", with the exception, that the year in the ntuples will be `2017v2` and L1Seeds will not be stored

PR #1419, opened 2019-08-29
new round of xml files for 2017 and 2018

pgunnell:

- Don't want to run any of this?  [ci skip]


raggleton:
Hi Paolo, could you please add the event counts to these XML files please? Thanks

PR #1425, opened 2019-09-16
[102X] Update 2017 MC Pileup histogram

raggleton:
Using the official mixing parameters in 10_2_10 for 2017:

```
makeMCPileupHist.py SimGeneral.MixingModule.mix_2017_25ns_WinterMC_PUScenarioV1_PoissonOOTPU_cfi
```

100 bins from 0 to 100

No big spike at 0 any more, should be suitable for use with 2017v2

[ci skip]


raggleton:
@abenecke this is the same as the 2017 MC pileup file you use, right?

abenecke:
Yes that is the one we use.

PR #1427, opened 2019-09-19
[102v2] Lower DeepBoosted min pt cut

raggleton:
Testing to see if it makes an effect (maybe not even needed since everyone cuts on pt anyway), don't merge yet


raggleton:
OK so there are some jets that were failing the cut before (gone from -10 -> [0,1]). But not sure it even matters, since analyses won't use that lower pT jets (plus might not even be valid in that range). It would also slightly increase runtime, just because more jets have to be passed through the deepboosted machinery.

Also because of me forgetting to update the branch name in install.sh, this is actually testing the V01 training.

PR #1429, opened 2019-09-24
[102X] Add photon selection and cleaning

karavdin:
Introduce tools for photon selection and cleaning in simialar fashion as done for muons and electrons.


raggleton:
Failure due to some jobs being killed (not sure why), rerunning them

PR #1435, opened 2019-09-29
[102X] Added 2018 JEC V19 and JER-SF V7

IreneZoi:
[only compile]

see https://hypernews.cern.ch/HyperNews/CMS/get/jes/898.html


raggleton:
@karavdin can we also clean up the "temporary" 2018 sets? v4, 7, 10, 13h, and the older official set v8?

karavdin:
@raggleton, I would leave v8 for some cross checks, but other temporary versions can be removed.

raggleton:
So the new JER SF are no longer simple values for each eta bin, they're now pt-dependent for 2.3  |eta| < 3.1. So we will have to now have machinery to parse the new txt files, as well as keep the old style for older versions (for now at least).

I'm also not sure where the V7 SF you added came from...can't see them in the txt files or wiki?

IreneZoi:
This is an example of where I found the SF https://github.com/cms-jet/JRDatabase/blob/master/textFiles/Autumn18_V7_MC/Source.txt

raggleton:
Ah I see. I was under the impression we use https://github.com/cms-jet/JRDatabase/blob/master/textFiles/Autumn18_V7_MC/Autumn18_V7_MC_SF_AK4PFchs.txt instead. 
Although, even in e.g the 0 - 0.522 eta bin where there's no pt dependence, the numbers don't agree between my file and Source.txt. So no idea where Source.txt plays a role? @karavdin , do you know if we should be using the Source.txt file?

karavdin:
Hi @raggleton, @IreneZoi, apologies for late reply. 
I have no idea the source of "Source.txt" and how it should be used. We usually copy the values from table on [twiki page](https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetResolution#Smearing_procedures). But this is not provided any more. 
The numbers we would need are stored in following files:
[ABC](https://github.com/cms-jet/JRDatabase/blob/master/textFiles/Autumn18_RunABC_V7_MC/Autumn18_RunABC_V7_MC_SF_AK4PFchs.txt), [D](https://github.com/cms-jet/JRDatabase/blob/master/textFiles/Autumn18_RunD_V7_MC/Autumn18_RunD_V7_MC_SF_AK4PFchs.txt), [ABCD](https://github.com/cms-jet/JRDatabase/blob/master/textFiles/Autumn18_V7_MC/Autumn18_V7_MC_SF_AK4PFchs.txt).

There is also original JER SFs version (without pt dependence) available in case we need something close to V7, which will work with our code: see txt files linked to the presentation [here](https://indico.cern.ch/event/837707/#5-dijet-jer-sf-update-for-2018).

**Current "SF_13TeV_Autumn18_V7" merged in this PR have different values in respect to official Autumn18_V7 JER SFs and should not be used.**

This is good motivation for us to extacrt/update respective CMSSW  module and use official txt files to apply JER smearing rather than copy and paste values.

PR #1437, opened 2019-10-01
[102X_v2] add 2017 DYJetsToLL xml files

froehlia:
[ci skip]


raggleton:
Just FYI, these were made with the RunII_102X_v2 branch with the new DeepBoosted training. I asked Alex to use this since we will need to process it anyway, and it didn't seem worth doing v1 and then v2 soon after. Because this is still compatible with the v1 branch, it may also be useful for everyone else, so can go into this branch.

I am wondering though if we should put such datasets in a new `RunII_102X_v2` directory in `common/datasets` to make this clear.

karavdin:
Hi @froehlia, @raggleton 
Please store samples created with branch RunII_102X_v2 in dedicated directory, this would make it a bit easier to track.

froehlia:
Sorry it took so long. Files are now in the correct directory

PR #1440, opened 2019-10-08
[102X] Reorder CommonModule process so returns go last

raggleton:
This ensures that all modules are always run, irrespective of the modules that return.

This fixes bugs where the user ignored the return value, but expected (fairly) that other modules were run.

This also updates the ExampleModule to show that users should be using the return bool

[only compile]

### DIFF
--- a/common/src/CommonModules.cxx
+++ b/common/src/CommonModules.cxx
@@ -168,14 +168,9 @@ bool CommonModules::process(uhh2::Event & event){
   if(!init_done){
     throw runtime_error("CommonModules::init not called (has to be called in AnalysisModule constructor)");
   }
-  if(event.isRealData && lumisel){
-    if(!lumi_selection->passes(event)) return false;
-  }
+
   for(auto & m : modules){
     m->process(event);
-  }
-  if(metfilters){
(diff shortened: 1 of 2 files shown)


reimersa:
This PR causes CommonModules to crash if running on data of a run not contained in the list of  runs from PdmV, which excludes some that aren't certified anyway (see https://github.com/UHH2/UHH2/blob/RunII_102X_v1/common/include/Utils.h#L135#L169 and https://twiki.cern.ch/twiki/bin/view/CMS/PdmV2016Analysis#DATA). 

The JSON is not checked until the very end, which results in RunSwitchers for JEC and JER to try and set up a corrector for a run number that is not actually contained in the JSON. An example for such an error is: "Message: RunSwitcher cannot handle run number 275656 for year 2016"

Right now, everyone would have to run the LumiSelection themselves before CommonModules. That doesn't seem to be the point of having it included in CommonModules as well?

raggleton:
Hmm good point. I originally moved it to the end so that every object gets treated, but as you say this breaks for JECs. Will move it back to the beginning in a new PR. I think the crucial thing is that people use the bool that it returns and actually veto their events.

PR #1447, opened 2019-10-15
[102v2] Use group storage area instead of individual user areas for CRAB jobs

raggleton:
This will change the default to write to the group storage area instead of the user's area. 

Still left in as a comment though in case of private production, etc.

It might be good to still retain the person's username in the request name though, for future reference? However with only 100 characters for the request name, that might be tricky. Or maybe we should just rely on the google sheet for looking that up. 
Thoughts? 

[ci skip]


karavdin:
google sheet was quite useful so far and main source of such kind information. Most of people don't know each other usernames  any way.

PR #1448, opened 2019-10-22
[102X] 2016 legacy JSON file

reimersa:
Added 2016 legacy JSON file


raggleton:
The failure was because it relies on CRAB API to do some tests, and this was broken

PR #1450, opened 2019-10-24
[102X] Added 2016 muon and electron SFs

reimersa:
Updated lumi-weighted average of muon SFs for 2016 legacy.
Added electron reconstruction SFs for 2016.


raggleton:
Stopped the tests because these files aren't tested (you can use `[ci skip]` in your PR message)

PR #1453, opened 2019-10-25
Skip scale variation if less than 9 systweights exist

reimersa:
[only compile]

Sometimes, some buggy events would have less than 9 systweights, although the sample generally has sufficiently many. Such events are now skipped.

### DIFF
--- a/common/src/MCWeight.cxx
+++ b/common/src/MCWeight.cxx
@@ -239,8 +239,8 @@ bool MCScaleVariation::process(Event & event){
   }
 
   try{
-    // Set weights to 1 if no genInfo exists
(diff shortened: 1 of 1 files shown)


raggleton:
Is this a bug/"feature" in the MINIAOD then for specific events in a sample? If you have an example dataset it would be good to know for future reference

reimersa:
For me that happened in at least two event in:
/ZZTo2L2Nu_13TeV_powheg_pythia8*/RunIISummer16MiniAODv3-PUMoriond17_94X_mcRun2_asymptotic_v3-v2/MINIAODSIM

Didn't happen in any other 2016v3 events I looked at so far...

PR #1461, opened 2019-11-21
[102X] v2 added QCD pT binnend sample with new training

abenecke:
[ci skip]


raggleton:
Just for future reference, these files are in your "RunII_102X_v1" dir on your T2 area (because the crab script hadn't yet been updated to the group area), so when it comes to cleaning up we should be a tad careful. At some point these should be moved to the group area anyway

PR #1484, opened 2020-01-28
[102Xv1] Clone UHH2-datasets repo as part of install script

raggleton:
Clone the UHH2-datasets repo as part of the standard install

(Probably should remove the common/datasets dir at some point, but may cause issues for people who havent pulled in a while?)

[only compile]


raggleton:
Failed because the `/cvmfs/cms.cern.ch/crab3/crab_slc6.sh` script to setup CRAB no longer exists...investigating where it's got to

PR #1487, opened 2020-02-10
[102X_v1] New version (replacement) of MCCSVv2ShapeSystematic module

MatthiesC:

Replace the old MCCSVv2ShapeSystematic module with a new version that also runs with DeepJet and DeepCSV. The .csv files with the input scale factors have to be downloaded separately and are not included in this PR


raggleton:
So the build worked, but `kinit` is failing. Needs fixing from my side, might be due to https://cern.service-now.com/service-portal/view-outage.do?n=OTG0054616. Or my credentials got updated.

raggleton:
Can someone else who uses b-tagging have a look? Not experienced with this bit of code or its usage, so could e.g. @froehlia check it please?

froehlia:
I'm not an expert with discriminant reweighting. I'm just using the normal scale factors for the standard working points. 
Looking at the code, this looks fine. Everything seems to do the same as before, with the only changes beeing a lot of renames and changing to use the BTag class instead of CSVv2.

PR #1490, opened 2020-03-16
Neural Network deployment

anmalara:
[102X] (But once tested, one could cherry-pick to [106X]

Add module base for NN applications.
Add example for user.
Modified MakeFile and Buildfile in UHH and examples folder to make possible to run TensorFlow software (Many thanks to @pgunnell and @raggleton for the help).
Hoping this compile out of the box, please test for different use cases (e.g. @finnlabe, @kdeleo @MatthiesC )

 [only compile]


anmalara:
Added Eigen package. This is extra, but very useful for liner algebra for NN.

[only compile]

raggleton:
CMSSW_10_2_X already has the eigen package available - can you use that if possible? Would rather avoid having to maintain it ourselves (and adding 1K extra files!)

raggleton:
BTW if you want to "undo" that last commit, it would be better in this case to do a git reset, and then force push (yes I know bad in theory, but no one else is using your fork's branch) to entirely remove it from history, otherwise we still end up inflating .git. 

Be warned that git reset _can_ remove uncommitted changes, depending on what sort of reset you do. Have a look at https://sethrobertson.github.io/GitFixUm/fixup.html#committed, or ping me for help.

common/python/__init__.py
```diff
@@ -0,0 +1,3 @@
+#Automatically created by SCRAM
```
raggleton: Can you remove this file please? It's autogenerated by scram, and having it in may cause breakages later

anmalara:
I added it following the example in core/python. Isn't it the same in principle?

raggleton:
I don't see an `__init__.py` in core/python? In either case, it's auto generated and shouldn't be added here

anmalara:
Done. No need to re-run the test. I guess. If you're fine, we can merge it.
Thanks for the help

PR #1491, opened 2020-03-18
Bug fix crash in MCPileupReweight

anmalara:
Don't know why this was still not fixed. I remember we did the same change long time ago.
It prevents the crash of the code if empty bin is found.

[only compile]

### DIFF
--- a/common/src/MCWeight.cxx
+++ b/common/src/MCWeight.cxx
@@ -184,6 +184,8 @@ bool MCPileupReweight::process(Event &event){
       cout << "WARNING no value in MC hist for trueNumInteractions = " << trueNumInteractions;
       cout << " Setting event weight_pu to 0" << endl;
       event.set(h_pu_weight_, 0.f);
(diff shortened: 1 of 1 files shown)


raggleton:
Yeah, I seem to recall we fixed a similar bug a while back... tbh the best way to handle this sort of could-return-prematurely behaviour, is to do the `event.set(handle, 0)` right at the start of the method, before anything can `return`, and then later call `event.set` again if it gets to the right values. This way it was at least set once to some default value, and we don't have to remember to add it in every time we want to do `return false`

raggleton:
Now it's saying there are merge conflicts...

anmalara:
it’s an indentation problem. I saw that only online but not on my terminal.
I’m in a meeting but I’ll fix it soon. Sorry about that.

anmalara:
Ok. Finally I fixed it. Sorry for the many attempts. I think we can merge it now.

PR #1494, opened 2020-03-30
Code for Hotmaps + UL updates

anmalara:
[only compile]

List of updates:

- Removed old hotmaps
- Added new ones for full Run2 + UL2017
- Expanded user case for JEC: now different level of correction can be used. Not interesting for common analyser but for JERC people
- Added code to run on hotmaps. will be updated because the recommendations are work in progress
- Added NoLeptonInJet ID
- Expanded JEC class for UL (17 and beyond I hope)
- Bonus: functions in Utils to retrieve the vector of run per year.


raggleton:
Since this modifies C++ files, can you please change your original PR message to include [only compile], then we can run it through the CI

PR #1495, opened 2020-04-02
[102X] fix bug in MCPileupReweight

froehlia:
[only compile]
fix a bug, where MCPileupReweight crashed, when the module was ran, without using up/down variation files


froehlia:
sorry, I mixed up some variable names. Should be correct now

PR #1508, opened 2020-05-06
Run ii 102 x v2 -- Importing changes from 102xv1

anmalara:
[only compile]

Couldn't find the relative commits in 102xv1. Copied files manually.
Added JER correction for HOTVR in the GenericSubJetCorrector class.
Minor bug fix for the NoLeptonInJet class (forgot the parameter was a boost optional).

### DIFF
--- a/JetMETObjects/README
+++ b/JetMETObjects/README
@@ -4,3 +4,7 @@ Slight changes have been made to enable stand-alone compilation and to avoid nam
 In particular, the Workaround.h file contains some code emulating cmssw edm::LogError to
 minimize required changes in other places of the code.
 
+The JetResolution class is from JetMETCorrections/Modules (not the one in CondFormats/JetMETObject!),
+whilst the JetResolutionObject class is from CondFormats/JetMETObjects
+
+In future, it would be good to just cms-addpkg instead of maintaining our own copies.
--- a/JetMETObjects/interface/JetResolution.h
+++ b/JetMETObjects/interface/JetResolution.h
@@ -1,0 +1,58 @@
+#ifndef JetResolution_h
+#define JetResolution_h
+
+// In standalone mode, no reference to CMSSW exists, so the only way to retrieve resolutions and scale factors are from text files.
+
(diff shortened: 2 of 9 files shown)


raggleton:
I'm not so keen on adding JER smearing & the new HOTVR-mass calculation to `GenericSubJetCorrector` - like the other JetCorrectors, it would be better if it only had one purpose, doing the JECs. Otherwise it ends up inconsistent with the other corrector classes. I'd rather the subjet JER smearing be done in a separate module or as an option in `GenericJetResolutionSmearer` (which also does the TopJets already).

I'm also a bit confused by the HOTVR mass calculation bit - (a) I don't understand what effect it has, since the TopJet collection is `const`, (b) how it is HOTVR-specific.
If it's just recalculating the mass from the subjets, I would probably recommend a TopJet method like `set_softdropmass_from_subjets()`, or `softdropmass_from_subjets()` - both have been on the TODO list for a while. Or modify `SoftDropMassCalculator` to allow the user to choose corrected or uncorrected subjets.

anmalara:
Well, In partially I agree with your concerns. The idea I had was the following:

if you have an HOTVR jet you correct the subjets with JEC and JER and then recalculate the mass. This re calculation was missing before. I thought of adding the JER in the same place as JEC just as a matter of simplicity. it's the same kind of loop and should be applied in both cases.  So, even though you might loose the "naming convention", we could find a smart way to name it, just not to have to maintain the same "corrections" in two different places. 

For the other questions, not so sure about a). I think I checked that the change was effective but I can crosscheck once more. b) it's not hotter specific I think. Not so sure what one should do with the normal AK8 topjets, but I think for hotvr it was needed. I'm fine with the suggestion on the method, but I'm not sure then how we should implement if, and if it has any impact on the rest of the ntuple production somehow. I also thought that the SoftDropMassCalculator has a different idea behind, nothing related to JEC/JER/SD...

anmalara:
What about this one? Any follow-up?

PR #1513, opened 2020-06-03
delete 102X_v1 2017 TstarTstar samples

finnlabe:

- Don't want to run any of this? (e.g. only XML files) Remove the @: [ci  skip]

These samples were reprocessed and are now available in 102X_v1.


anmalara:
@raggleton I think those files are into Anastasia folder. Can you remove them? Same for the other PR in uhh2-dataset.

raggleton:
I'm afraid I can't remove the files in someone else's user directory, only those in the group directory

anmalara:
Do we leave there then? I guess her area will be deleted at some point.

raggleton:
Yeah, I'm assuming as a signal sample it's not that big anyway

PR #1515, opened 2020-06-17
[102X] Add checking for 'TTTo' in ttbar file name in TopPtReweight

froehlia:
[only @ compile]

added another check to the list of possible names for ttbar samples.
In 2017 and 2018 the ttbar samples usually are named "TTTo[...]", which was not captured before.
Now it checks if the name starts with "TTTo"


raggleton:
Failure due to a credentials issue - PR is fine otherwise

PR #1519, opened 2020-07-03
[106X] updated to v14 with CMSSW 10_6_13

abenecke:
I have changed the install script to CMSSW_10_6_13 which makes it easier to use the new default PUPPI tune v14 for UL production.

I also added the UpdatePuppiTuneV14 from
https://github.com/cms-sw/cmssw/blob/CMSSW_10_6_13/CommonTools/PileupAlgos/python/customizePuppiTune_cff.py

which updates our PUPPI jet collections with the latest tune.

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -1173,6 +1173,10 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
     )
     task.add(process.rekeyPackedPatJetsAk8PuppiJets)
 
+    #### update PUPPI to v14
+    from CommonTools.PileupAlgos.customizePuppiTune_cff import UpdatePuppiTuneV14
+    UpdatePuppiTuneV14(process, not useData)
+
     # Update DeepBoosted training to V2 for everything but 2016v2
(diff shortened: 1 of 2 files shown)


raggleton:
By default this hasn't triggered the full cmsRun chain, since I don't have any configs/test files in place. But it would be good for someone (e.g. @anmalara) to try it on a small sample of events and check that all the PUPPI-related collections show a change

abenecke:
I tested PUPPI pT and charged multiplicity. Both show a change between the original and the new tune. 

Screenshot 2020-07-03 at 19 28 19
Screenshot 2020-07-03 at 19 31 11

raggleton:
Nice, very clear it's working. 
I think the other things interesting to check before/after are: the same for the AK8 PUPPI jets, and to see if the deepX & deep flavour discriminator values change; and to also plot PUPPI MET.

abenecke:
image
image
image

I attached the comparison of AK8 jets pT and charged multiplicity and PUPPI MET. Each quantity changes between the old and the new tune.

raggleton:
Some other changes in this PR that may affect/need to be fixed in future:

- jetsAk4Puppi loses `partonFlavour`
- jetsAk4Puppi loses deepCSV (although it's not supported for PUPPI jets anyway)

Issue #1521, opened 2020-07-16
Issue locating JER files


jngadiub:
I installed and compiled UHH2 following the recipe [here](https://github.com/UHH2/UHH2/wiki/Installing,-Compiling,-Ntuples-(RunII-2016,17,18-datasets-in-CMSSW_10_2_X-v2)):

```
ssh -XY lxplus6.cern.ch
mkdir UHH2
cd UHH2
wget https://raw.githubusercontent.com/UHH2/UHH2/RunII_102X_v2/scripts/install.sh
source install.sh
cmsenv
cd ../../../SFrame
source setup.sh
make -j4
cd ../CMSSW_10_2_16/src/UHH2
make -j9
```

Then I download my [analysis code](https://github.com/jngadiub/UHHNtupleConverter.git) and compile it. When running it I get the following error:

```
 ( FATAL )  sframe_main        : app/sframe_main.cxx:70 (int main(int, char**)): STD exception caught
 ( FATAL )  sframe_main        : app/sframe_main.cxx:71 (int main(int, char**)): Message: Registry: exception while trying to build type 'UHHNtupleConverterModule': Could not locate file '2016/Summer16_25nsV1_MC_PtResolution_AK8PFchs.txt' in $CMSSW_BASE/src/{UHH2,} or $SFRAME_DIR/{UHH2,}.
 ( FATAL )  sframe_main        : app/sframe_main.cxx:72 (int main(int, char**)): --> Stopping execution
```

These files do exist though under `$CMSSW_BASE/src/UHH2/common/data` but they are not picked up.
The JER files are loaded automatically/internally from the year and collection (i.e., I do not indicate in my analysis code a custom path). Also note, that this issue does not appear when basing on the `RunII_102X_v1` branch. In fact, I noticed a difference with the `RunII_102X_v2` one:

The `RunII_102X_v2` branch comments [here](https://github.com/UHH2/UHH2/blob/RunII_102X_v2/common/include/JetCorrections.h#L276) that the path `common/data` is automatically attached  but this does not seem to be the case as you can see in the [JetCorrections.cxx#L915](https://github.com/UHH2/UHH2/blob/RunII_102X_v2/common/src/JetCorrections.cxx#L915). In contrast, this was happening in the `RunII_102X_v1` branch, i.e. at [JetCorrections.cxx#L913](https://github.com/UHH2/UHH2/blob/RunII_102X_v1/common/src/JetCorrections.cxx#L913).

Did anybody else experience this problem?

[...]

PR #1522, opened 2020-07-16
fixed path for JER files

jngadiub:


anmalara:
After discussion in  #1521:

I noticed that the problem is in the default constructor for JetResolutionSmearer, rather than in the GenericJetResolutionSmearer.
I'd rather adjust the path for the resolution file, as for the SF files. In this way the new idea are still available and the module is also backward compatible.

Can you please:
- don't prepend common/data in GenericJetResolutionSmearer
- fix the path in JetResolutionSmearer 
- add a check & message to locate_file() that goes into the JR ctor ("JRDatabase/textFiles/"): if it can't find it, then tell the user that they may need to prepend "common/data" ?

Here the code location.
https://github.com/UHH2/UHH2/blob/RunII_102X_v2/common/src/JetCorrections.cxx##L860
https://github.com/UHH2/UHH2/blob/RunII_102X_v2/common/src/Utils.cxx##L54

raggleton:
Fixed more generically using JRDatabase in #1536

PR #1523, opened 2020-07-22
Run ii 102 x v2 -- Update Muon and Ele SF modules (dummy bins and hist name as argument)

anmalara:
[only compile]


anmalara:
Moved the default hist name for EleSf module into the constructor. Now one can plug other names too.
In Muon and Ele SF modules added check for "dummy" bins. As mentioned in twikies, if sf=1 and err=0, consider adjacent bins.

PR #1534, opened 2020-09-22
[102X] Added a DetectorCleaning module.

tom-so:
Include a module which removes events in case electrons/muons/jets are found in an area specified by an eta and phi interval.
Includes module HEMCleanerSelection with default values for the HEM15/16 issues.

### DIFF
--- a/common/include/DetectorCleaning.h
+++ b/common/include/DetectorCleaning.h
@@ -1,0 +1,46 @@
+#pragma once
+
+#include "UHH2/core/include/Event.h"
+#include "UHH2/core/include/Selection.h"
+
+#include "UHH2/common/include/Utils.h"
+
+using namespace uhh2;
+using namespace std;
+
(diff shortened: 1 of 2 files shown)


raggleton:
Testing pipeline broke due to unrelated reason; code looks & compiles fine.

BTW in future for an AnalysisModule, you can add `[only compile]` to your original Pull Request message: this tells the testing pipeline to only compile your code. Otherwise it tries to run the whole pipeline that makes ntuples (which is much longer, and doesn't test any of your changes in the `common` directory)

@anmalara Is this needed in Ultra-Legacy samples? I understood that there was some HEM mitigation, but not sure if same veto still recommended

anmalara:
afain, for UL the recommendation is to use the 2D jet veto maps (still to implement from my side) in which one can choose the one with HEM issues already included. I'd still do the cherry-pick into 106X, since it's a generic module and might me used in the future.

PR #1537, opened 2020-10-26
[106X] JER file locating method

stalbrec:
Port from 102XV2 (#1536), so:

> This adds another method to JERFiles namespace in JetCorrectionSets for locating MC JER files from the JRDatabase in similar fashion as done for the JEC files.

Also updates the recommended JER file version for UL18 (from the mentioned [TWiki)]( https://twiki.cern.ch/twiki/bin/view/CMS/JetResolution#JER_Scaling_factors_and_Uncertai) - is this correct?

I noticed there are some checks on whether the JER-filenames contain `JRDatabase/textFiles/` (in [JetCorrections.cxx#L625](https://github.com/UHH2/UHH2/blob/0210ca1861eff68dc71c9b9e0041b5dde42176f5/common/src/JetCorrections.cxx#L625) ). These are now redundant, correct? Should i remove those two lines and add a commit to this PR?

[only compile]

### DIFF
--- a/common/include/JetCorrectionSets.h
+++ b/common/include/JetCorrectionSets.h
@@ -9,6 +9,17 @@
 namespace JERFiles{
 
   /**
+   * JERPathString returns the path of a JER file from JRDatabase for a given 
+   * tag, 
+   * jetCollection,
+   * type (i.e. SF,PtResolution, etc.).
+   * To be used for example in JetResolutionSmearer or when constructing a GenericJetResolutionSmearer
+   */
+  const std::string JERPathStringMC(const std::string & tag,
+                                  const std::string & jetCollection,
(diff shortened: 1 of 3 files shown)


raggleton:
> Also updates the recommended JER file version for UL18 (from the mentioned TWiki) - is this correct?

@anmalara can confirm, but let's stick with the Twiki recommendations for the average user

> I noticed there are some checks on whether the JER-filenames contain JRDatabase/textFiles/ (in JetCorrections.cxx#L625 ). These are now redundant, correct? Should i remove those two lines and add a commit to this PR?

Yes, that was probably a first attempt at solving the directory/filename issue. So if you rm those lines, push another commit, we can rerun the compilation

anmalara:
That's correct for the moment.

PR #1538, opened 2020-11-04
[102X] cherry-pick: Fix order in CommonModules so exits first

froehlia:
This means that invalid runs won't be passed onto later
modules, e.g. JECs, that depend on run number.

The previous idea about being 'safer' was wrong -
only by using the return value can the analyzer
properly handle things.

[only compile]


raggleton:
Sorry for being so late in checking this. Thanks for spotting this, I should have cherry picked this when I did #1449 , not sure how it escaped

PR #1542, opened 2020-11-17
[102Xv2] Use slc7 SCRAM_ARCH for EL7 machines

raggleton:
Allows the 102X branches to be used on EL7 machines properly, without resorting to singularity.
Requires a slight increase in CMSSW version as `10_2_16` doesn't exist for `slc7_*`. (I don't want to jump versions to the latest, since I'm not sure if there are important differences in there)

The install script now tries to determine machine's architecture, and then sets `SCRAM_ARCH` accordingly.

[only compile]

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -84,31 +84,21 @@ setupFastjet() {
 }
 
 
-checkArch() {
-    # Check if this machine is compatible, because at CERN the default is EL7,
-    # whereas we need SL6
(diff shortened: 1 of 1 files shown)


raggleton:
Failures were due to adding a new build stage to testing pipeline to test both SL6 and EL7 setups. 
Now works fine, using `/cvmfs/cms.cern.ch/common/cmsos` to do the hard work of determining the arch (more universal than just e.g. `uname -a`)

PR #1544, opened 2021-01-14
[106X] Remove JetBTagInfo class

raggleton:
Removed since it has a lot of unused memory assigned to it in each Jet collection (see e.g. https://uhh2-integration.web.cern.ch/UHH2integration/test1535/mc_UL18_TTSemiLeptonic.html#sizejetsAk4Puppi) and was only previously used for subjet b-tagging SFs/calibrations (I think?)

Removed the header, and references to it in classes.h/classes_def.xml, to tidy up.

This should save a fair amount of space (5.9 -> 5MB on 500 TTBar events using ntuplewriter_mc_UL18.py, let's see how it does on other samples)


raggleton:
Looks like removing the `JetBTagInfo` attribute can save a fair amount of space, ~15 - 20%. Unless anyone actually needs this info, I'd be strongly in favour of removing it.

(The change in timings is uncorrelated with this change, and is just the natural variation across worker nodes)

mschrode:
The object/info was used in the past for btag calibration purposes, but it has not been used since a while. Conclusion from the user side is that it can be removed.

PR #1547, opened 2021-02-08
new UL18 jec tag and version

MatthiesC:

Updated UL18 jec tag and version in CommonModules

### DIFF
--- a/common/src/CommonModules.cxx
+++ b/common/src/CommonModules.cxx
@@ -34,9 +34,8 @@ CommonModules::CommonModules(){
   jec_tag_UL17 = "Summer19UL17";
   jec_ver_UL17 = "5";
 
-  // FIXME: update when official set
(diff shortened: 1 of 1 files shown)


mschrode:
Hi @MatthiesC, one question: do you expect any of the test distributions (see above) to change due to the updated JEC tags? Naively, I would have thought that some of the jet distributions become different, but they are either exactly the same compared to the reference version or not filled at all in both versions (which is unrelated, I think, eg due to missing btaginfo etc.)

anmalara:
@mschrode  The tag is used only on fly and I believe that the ntuple step refers to the production, for which we use the GT.
I think @MatthiesC forgot to ask for a compilation-only CI.

mschrode:
Ah OK, makes perfect sense, thanks @anmalara for the clarification

PR #1549, opened 2021-02-22
Update HEM issue module [102X]

anmalara:
[only compile]
Extended to module to do a cleaning on jets and top jets separately.

### DIFF
--- a/common/include/DetectorCleaning.h
+++ b/common/include/DetectorCleaning.h
@@ -10,7 +10,7 @@ using namespace std;
 
 class EtaPhiEventCleaner: public uhh2::Selection {
 public:
-  EtaPhiEventCleaner(uhh2::Context& ctx, int min_run, int max_run, float min_eta, float max_eta, float min_phi, float max_phi, std::string jetCollection = "topjets", bool doJets = true, bool doElectrons = true, bool doMuons = true);
+  EtaPhiEventCleaner(uhh2::Context& ctx, int min_run, int max_run, float min_eta, float max_eta, float min_phi, float max_phi, std::string JetCollection = "jets", std::string TopJetCollection = "topjets", bool doElectrons = true, bool doMuons = true);
   virtual bool passes(const uhh2::Event& event) override;
 private:
   int min_run;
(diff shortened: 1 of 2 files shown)


mschrode:
Hi @anmalara , in general, is it possible and does it make sense to pass the strings for Jet/TopJetCollection as "const reference" instead of "by value"? I don't think it makes a huge difference but should be a tiny bit faster and less memory consuming. Not sure though how much changes it would require down the line, since the argument gets passed on?

anmalara:
Right. It makes sense. Should work without much trouble

mschrode:
Ah, no, sorry, I meant passing the const reference, e.g. 

std::string JetCollection --> const std::string& JetCollection

so with the "&". This will pass the reference instead of creating a new copy of the object passed as argument. This is only useful for non-basic types such as std::stding; for float etc. it does not gain.

MatthiesC:
@mschrode Since these only are arguments to the constructor of a uhh2::Selection daughter class, they will be passed only once per analysis cycle, not per event. Making them passed-by-reference or not thus won't make a noticable difference. It would only be really important/noticable for member functions of a uhh2::Selection (or uhh2::AnalysisModule) which would be called per event (but most of the time, one should stick with the uhh2::Selection::passes(const uhh2::Event&) and uhh2::AnalysisModule::process(uhh2::Event&) routines regardless and pass all needed parameters to the class via its constructor, as done by @anmalara ).
Nevertheless, I support the use of pass-by-reference wherever possible and needed, of course

anmalara:
@mschrode, you're right! I forgot to add the "&" in the search-replace function, and I didn't notice the bug, since the compiler didn't complain. I also moved from "static" to "inline static", otherwise the reference doesn't work.

MatthiesC:
> 
> 
> @mschrode, you're right! I forgot to add the "&" in the search-replace function, and I didn't notice the bug, since the compiler didn't complain. I also moved from "static" to "inline static", otherwise the reference doesn't work.

@anmalara  Why do you consider non-passed-by-reference arguments to be a bug?

anmalara:
Because that's not what I wanted to do! ;)

mschrode:
Hi @anmalara @MatthiesC , thanks for looking into it and sorry for all the noise/extra work for a probably rather academic question... (Just to be super picky: for most basic types like int, float, etc. passing by value is actually preferred over by reference, see eg [this discussion](https://stackoverflow.com/questions/14013139/is-it-counter-productive-to-pass-primitive-types-by-reference) - but it won't matter so no need I think to change anything.)

If you agree, I'll merge the PR now.

anmalara:
@mschrode I didn't know that for basic types was not preferred. Good to know. I think now we agree. Right?

mschrode:
Thanks, merged!

PR #1550, opened 2021-02-23
[106X] Using Puppi v15 in Ntuple generator (re-installation of UHH2 required since newer CMSSW version is needed)

MatthiesC:

**Updates:**
- Using Puppi v15 in Ntuple generator, following instructions from https://twiki.cern.ch/twiki/bin/viewauth/CMS/PUPPI#Recipe_for_106X
I will attach sanity-check plots in a separate comment to this PR
- This recipe requires CMSSW_10_6_>=17, thus updating the UHH2 install scripts
- This new CMSSW version also requires a further update: ONNXRuntime replaces MXNet (frameworks used for DeepBoosted stuff), see e. g. https://github.com/cms-sw/cmssw/pull/30123 and https://github.com/cms-sw/cmssw/pull/31037
However, this comes with a change of the DeepBoosted outputs, see plot in following comment to this PR (I am not sure why, most likely it is due to a change of the training version)
- Replaced source files for testing of UL17 ntuplewriters (old ones do not exist anymore)

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -1187,23 +1187,21 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
     )
     task.add(process.rekeyPackedPatJetsAk8PuppiJets)
 
-    #### update PUPPI to v14
-    from CommonTools.PileupAlgos.customizePuppiTune_cff import UpdatePuppiTuneV14
-    UpdatePuppiTuneV14(process, not useData)
+    #### update PUPPI to v15
+    from CommonTools.PileupAlgos.customizePuppiTune_cff import UpdatePuppiTuneV15
+    UpdatePuppiTuneV15(process, not useData)
 
     # Update DeepBoosted training to V2 for everything but 2016v2
     # Check https://twiki.cern.ch/twiki/bin/view/CMS/DeepAKXTagging for latest recommendations
     # e.g. 2018 specific training
     if (year != "2016v2"):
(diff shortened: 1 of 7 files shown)


MatthiesC:
As seen in these plots, the update from v14 to v15 has an effect

ak4puppi_pt
ak8puppi_pt
ak8puppiSoftdrop_pt
hotvrpuppi_pt
xconepuppi_pt

MatthiesC:
As seen in this exemplary plot, the change from MXNet to ONNXRuntime has an impact on the DeepBoosted output shapes, meaning that the change likely also comes with a different training. Does anyone know details?
For this plot here, I explicitly chose AK8 CHS jets in order to reject the hypothesis that the different shapes come from the new Puppi version (indeed, other distributions in CHS jets remain unaffected by the Puppi version change)

ak8chsSoftdrop_DeepBoosted_WvsQCD

MatthiesC:
> https://gitlab.cern.ch/raggleto/UHH2-integration/blob/test1550/scripts/cmsrun_jobs.py

I just noticed that this link is not available anymore (the link is given in the uhh2 bot messages above). Needs a fix

MatthiesC:
We verbally agreed on not merging this PR.
We want to stick with Puppi v14 for now, reasoning:
- Avoid confusions regarding which ntuples in UHH2 pnfs storage will have v14 or v15, and
- JEC and JER for UL17/18 have been determined with v14 so far.

Test source files for ntuplewriters for UL17 don't exist on pnfs/desy.de/cms/tier2/, but _maybe_ somewhere else, thus do not change them without further checks in that direction.
Last but not least, CMSSW_10_6_17 will not be the final UL branch. Thus, avoid updating it right now

PR #1554, opened 2021-02-26
Run ii 106 x v1

anmalara:

- Only want to compile, but don't make + test ntuples? (e.g. change `CommonModules.cxx`) Remove the @: [only compile]


MatthiesC:
Seems that there is a conflict. I already added the JetPFIDs for UL (alongside other minor changes regarding the LepVetoID being different for CHS and PUPPI in EOY 2018)

anmalara:
HI @MatthiesC. Yup. The problem is that I didn't want to pull the latest status since I'm working with files that have the Btag info. I though I cherry-picked all the important commits, but apparently I didn't. I hope I fixed it now

common/src/JetIds.cxx
```diff
@@ -682,60 +681,27 @@ bool JetPUid::operator()(const Jet & jet, const Event &ev) const{
...
-        h2HotExcl[year] = (TH2D*) file_map.Get(mapHistname.c_str());
-        h2HotExcl.at(year)->SetDirectory(0);
-        file_map.Close();
-      } catch (...){}
-
-      continue;
-    }
+  std::string year = "UL16postVFP";
+
+  TFile file_map(locate_file("/nfs/dust/cms/user/amalara/WorkingArea/UHH2_106X_v1_UL/CMSSW_10_6_13/src/UHH2/DiJetJERC/conf/hotjets-UL16.root").c_str());
```
MatthiesC: Please make this filepath user-independent

MatthiesC (changes requested):

MatthiesC:
> 
> 
> HI @MatthiesC. Yup. The problem is that I didn't want to pull the latest status since I'm working with files that have the Btag info. I though I cherry-picked all the important commits, but apparently I didn't. I hope I fixed it now

Ok. As I said in a meeting some days ago: I don't really know why it should be a problem using ntuples with the btag info together with the newest UHH2 setup without the btag info class. I only get warnings but no runtime errors or similar.

common/src/JetIds.cxx
anmalara: Thanks for pointing that out.
I thought for a bit if it made sense or not. I decided for not for the following reasons:
1) That file would eventually go into the JEC database. Right now it has only circulated "internally" for the JERC studies.
2) This module is not recommended for anyone in our group, but used for the JERC studies only. (Aka. Me and Alex at the moment)
3) I hope the recommendations on which file to use are going to be finalised in the next 2 weeks. Afterwards I will make sure this module is used properly.
4) Adding now and then remove a root file in a couple of weeks is not really nice for the git idea.

I had to modify the module otherwise the code was crushing, given the update. And it's also needed for me and Alex. I also though if I should comment the module at all. I hope I can fix it shortly.
 
Said that, I agree it's a bit ugly. Although I'm not leaving the group in the next two weeks.
What do you think? 

The other solution would be to leave this PR open for a couple of weeks. But then we should be careful of not making any merge conflict! I'm open for suggestions ;)

anmalara:
> > HI @MatthiesC. Yup. The problem is that I didn't want to pull the latest status since I'm working with files that have the Btag info. I though I cherry-picked all the important commits, but apparently I didn't. I hope I fixed it now
> 
> Ok. As I said in a meeting some days ago: I don't really know why it should be a problem using ntuples with the btag info together with the newest UHH2 setup without the btag info class. I only get warnings but no runtime errors or similar.

This is also true. Still, due to time-constrains I couldn't test this new setup, nor I wanted to mix it with my current setup. In any case, this is really not a problem, since I already solved the merge conflict ;)

common/src/JetIds.cxx
MatthiesC: Given your explanations, it's okay to just merge it now I think. But we should make sure to not forget updating that filepath in a few weeks once that hotzone map for UL16 is added in the JECDatabase (if I understand correctly). In order to make sure to not forget it, could you maybe add a std::cout or runtime warning to the module? Just an idea

mschrode: Hi @anmalara @MatthiesC do you want me to merge the PR like this or do you prefer to add a cout like suggested?

anmalara:
Sorry it took so long. I asked the conveners to add the veto maps into the repository. Now they are available and ready to be used. I implemented the latest recommendation for UL. EOY is not supported yet. In any case, this module is still "under construction". So might be in the future it will get updated.

PR #1557, opened 2021-03-26
[102X] neural network deployment for TF2 networks

finnlabe:
[ci skip]

A short python script analogous to common/python/prepare_for_deployment.py but for TensorFlow 2.
Let me know if there is anything that should be changed/added.


mschrode:
Hi @finnlabe , since this is a new file and since you guys already discussed a bit in mattermost, I'll merge it right away so ppl can use it. If we want to change or add sth, we can do a new PR

anmalara:
Can you do a cherry-pick to 106X?

PR #1558, opened 2021-04-06
[102X] Fixed string interpolation compilation error in TF2 deployment script

finnlabe:
[only compile]

I did not know that python files are compiled in UHH2. When I compiled UHH2 offline, i noticed an error in "prepare_for_deployment_TF2.py", which uses string interpolation that is not available in the used python version.

To avoid anyone stumbling over this compilation error, this PR fixes it (and now uses "only compile" to check whether compilation works).


finnlabe:
I commited to the same branch and added a second commit to the PR, which was not intended. To keep this clean, I will close the PR and do a new one. Sorry, did not know this git behavior.

PR #1561, opened 2021-04-14
[106X_v1] jecRunMap for UL16APV

apaasch:


mschrode:
Hi @apaasch , failing seems to be caused by the tarball size check in [makeCheckCrabTarball](https://github.com/UHH2/UHH2/blob/RunII_106X_v1/scripts/crab/makeCheckCrabTarball.py) The file hasn't been touched recently, possibly it is related to changes in the crabclient interface [UserTarball](https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/UserTarball.py) where we take the FILE_SIZE_LIMIT from. There have been some changes in that file, but I don't understand really what's going on

mschrode:
The same seemd to have happened in #1560

anmalara:
I think that happens in the 106 since I added some csv files for lumi stuff. They are ~20-30 MB each, so not too large. I would not remove them. Is it possible to mask these files? Or in general files that are not needed for crab jobs?

mschrode:
Ahh, that makes sense, that must be it!

mschrode:
Not sure if/how one can mask files for crab tarballs. I seem to vaguely recall it is possible...

anmalara:
It is possible, from the core/python/bsframe.py. The point is how to make sure we exclude the correct files. My knowledge stops where and how this tarball is used.

https://github.com/UHH2/UHH2/blob/RunII_102X_v2/core/python/bsframe.py#L564

mschrode:
I also need to confirm, but I think for the CI pipeline it is just to test the tarball size in fact, there are no crab jobs submitted

anmalara:
Do you know if it is also used when creating ntuples? That's the only time we really use crab

mschrode:
not really, no. But I would expect we can exclude the csv files - if they are not needed for the "regular" ntuple production, it would do no harm. Inversely, "crab submit" should fail in the current setup because the tarball size would be too large. (Still, of course, need to check/understand a bit better where/when the script is used.)

raggleton:
To help decipher this a bit: in the CI we run 2 relevant tests, (1) it creates a sandbox exactly as CRAB does (`makeCheckCrabTarball.py`), and (2) we need a similar tarred up version of sframe + CMSSW (incl. UHH2) for future steps in the gitlab CI pipeline (`tarballEntireRelease.py`).

(1) uses the settings in `scripts/crab/crab_template.py` to determine what gets included in the sandbox. It also does a check against the CRAB internal size limit, exactly as if you submitted CRAB jobs. At the moment this passes (although it's at 95MB out of a limit of 117MB, so it's getting close - I'd have a look at slimming it down if possible https://gitlab.cern.ch/mschrode/UHH2-integration/-/jobs/13208054#L4856)

Note that it automatically includes certain directories, which I don't think you can't opt out of including. Nor can you upload your own sandbox as far as I know.

(2) is just so we can easily setup SFrame + CMSSW + UHH2 in the following jobs (this is uploaded to gitlab as an "artifact"). Here we have to package up more stuff. Now, I do a size check here (a) to keep an eye that it hasn't ballooned massively, and (b) because gitlab has a limit on the artifact size of 100MB nominally: https://docs.gitlab.com/ee/user/admin_area/settings/continuous_integration.html#maximum-artifacts-size

(2) is the script that is failing: https://gitlab.cern.ch/mschrode/UHH2-integration/-/jobs/13208054#L4916

However, there are 2 unknowns here:
- this limit is administered by CERN, and isn't visible to us users. So it may be set at the default of 100 MB, or it may be different. Couldn't find CERN docs about this.
- job artifacts are compressed prior to being uploaded to gitlab. I guess the file size limit is on the _compressed_ artifact; however since the artifact we produce is already GZ compressed, I am not sure it will make a big difference. Might be worth trying? You can look at older pipeline's artifacts to compare compressed vs uncompressed sizes.

There is also a way to print out the size of things in a GZ tarball: https://gitlab.cern.ch/mschrode/UHH2-integration/-/blob/master/scripts/makeCheckCrabTarball.py#L79

You may want to add some sort of filter on the input to `tarballEntireRelease.py`, such that large unnecessary (for ntuple production) files are excluded.

mschrode:
Hi @raggleton thank you so much for helping out here! For some reason I did not manage to identify the script (2) that was actually causing the fail.

mschrode:
For (my) reference: I think this is fixed now with [this commit in UHH2-integration](https://gitlab.cern.ch/mschrode/UHH2-integration/-/commit/8008e51690ffe044cea4be5eb604eafecc929771), which removes the entire `UHH2/common/data` from the tarball that is packed for the gitlab CI. The commit adds a filtering function that we can adjust to a finer selection in the future if needed, but for now the files in `data` do not seem to be used in the automatic tests.

PR #1563, opened 2021-04-29
[106X] UL updates: renamed muon ID and ISO SF files; added SingleMuon trigger SF files

MatthiesC:

- Don't want to run any of this? (e.g. only XML files) Remove the @: [ci  skip]


mschrode:
Thanks @MatthiesC ! For the *renamed* files, are they imported somewhere in the repo, eg a config file, that needs adaption then as well? And for the *added* files, the same, do they need to be added to some config sequence in the repo?

mschrode:
Ah, one more question: how large are the added files?

MatthiesC:
Hi @mschrode! The largest file has a size of 104kB, all files have 1.1MB. The filepaths are defined on user-level, so there are no adjustments to be made to the main code.

The files are read-in by the class `MCMuonScaleFactor` in https://github.com/UHH2/UHH2/blob/RunII_106X_v1/common/include/MCWeight.h#L107

mschrode:
OK, I see, thanks!

PR #1564, opened 2021-04-30
[102X] Adding L1 Info for MC and Data

finnlabe:
Adding L1Muon, L1Tau and L1Etum collections and change L1Jets and L1EGamma so that they can also be stored for data.
Switching the default behavior so that they are not stored by default.

Compiled for me and looked good in tests, lets see what the bot says :)

### DIFF
--- a/core/include/Event.h
+++ b/core/include/Event.h
@@ -6,7 +6,7 @@
 namespace uhh2 {
 
 /** \brief The event container.
- * 
+ *
  * Based on the UHHanalysis' BaseCycleContainer. This class allows easy (direct) access to many frequently needed
  * objects, such as jets, electron, muons, etc. To store data beyond those directly declared data
  * members, use the GenericEvent methods, which allows settings and getting arbitrary data via handles.
@@ -14,7 +14,7 @@ namespace uhh2 {
 class Event: public GenericEvent {
 public:
   double weight; // Note: unlike the other members, the weight is not read from the input file and not written to the output file, but calculated on the fly.
-    
+
   int run;
   int luminosityBlock;
   long long event;
@@ -34,8 +34,8 @@ public:
   float beamspot_y0;
   float beamspot_z0;
   std::vector< PrimaryVertex >* pvs;
-  
-  std::vector< Electron >* electrons;  
+
+  std::vector< Electron >* electrons;
   std::vector< Muon >* muons;
   std::vector< Tau >* taus;
   std::vector< Photon >* photons;
@@ -46,6 +46,7 @@ public:
   MET* genmet;
   std::vector< L1EGamma>* L1EG_seeds;
   std::vector< L1Jet>* L1J_seeds;
+  std::vector< L1Muon>* L1M_seeds;
 
   GenInfo* genInfo;
   std::vector< GenTopJet >* gentopjets;
@@ -54,7 +55,7 @@ public:
   std::vector< PFParticle>* pfparticles;
 
   /** \brief Access to trigger results.
-   * 
+   *
    * Access to trigger results is treated differently from the other
    * event content in order to save disk space: In principle, the trigger
    * information saved here can be thought of as a map from string to bool,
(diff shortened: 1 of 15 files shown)


finnlabe:
Ah. Obviously, when not including L1 objects by default, they are then not included in the tests (this explains the two removed collections). Is there a way to test with `doL1seeds = true` instead?

mschrode:
right... :) I'll check to adjust the tests. I think it can make sense to always include it in the tests, also for future PRs, to make sure nothing silently breaks the L1-related entries

mschrode:
Hi @finnlabe , I kind of lost track here, sorry: did we want to merge this PR still for 102X? If so, is this up-to-date with your latest planned modifications (switch which variables to include etc.) or will there be another commit that we should wait for?

finnlabe:
Hi! I will push two additional commits today, doing two things:

- Fix inclusion of L1prescales to be done for all (data) samples, not steered by `doL1seeds` anymore
- rename `doL1seeds` to something like `doL1objects` for clarity

This will not change the results of the tests. The links above are broken for me, so maybe we need to run tests again. I have checked that most histograms are filled in the root files I generate, whereas they were shown as empty in the tests (cant check 100% as they are gone). I think we should look into the reason for this.

mschrode:
Hi @finnlabe , OK great, thanks for reminding me. 

Indeed, the individual links in the table are broken (as Steffen already informed me), I'm fixing this for new PRs. You can still access the control plots of this PR, the link given in the text above the table is correct: https://uhh2-integration-results.web.cern.ch/uhh2-integration-results/UHH2integration/test1564

finnlabe:
Hi @mschrode, that is good to know, thanks! I did check the plots for the new commit I just pushed (that should not change anything technically). I see some discrepancy between my plots and the one from the tests: Using, for example `ntuplewriter_mc_2018.py`, including TTToHadronic, L1Muon_seeds.charge() is empty in the tests, and shows expected behaviour (two bins, one at -1 and one at 1) for me. I assume the plotting ranges for the test histgrams are chosen automatically? (Otherwise that would explain why it looks empty, as all entries are outside of (0,1).)

mschrode:
Thanks @finnlabe ! I just re-triggered the tests with your new commit included. As for the plotting range, good question, will need to check this

finnlabe:
I just had a look into the new plots, they are very similar to the previous results. So there are still some discrepancies (again muon charge as an example). Let me know @mschrode if you find something regarding the plots ranges.

mschrode:
Hi @finnlabe , I think it is fixed [here](https://gitlab.cern.ch/mschrode/UHH2-integration/-/commit/fb786ca5a7fe916ab9edebbfedde719df7386ffe). I re-triggered the tests, let's see what happens... ;)

mschrode:
Looks different :)

finnlabe:
Yes I can confirm that this looks in line with what I saw offline. Great, so it really just was a plotting range problem :D
Now the last question is regarding the default setting: we kept it as `doL1TriggerObjects = true` for the tests, but decided we want it on `false` finally. Should there be some change to have it set to `true` for tests?

mschrode:
ah, right, thanks for reminding about the switch. I need to understand better how to set it to `True` only for the tests, but I think I can do that from the test workflow independently of "your" code. This I need to take a closer look still, though. I suggest the following:
1. for this PR in 102X, we put the switch to `False`. This means you need to change the commit still one more time, right?
2. you can then cherry-pick this PR to the 106X branch, also with the switch to `False`
3. in parallel, I'll see to adjust the test logic but we keep it independent of the PRs so that we can merge them for now

finnlabe:
See last commit: this changes the boolean to false by default. I will cherry-pick all commits in this PR to 106X later today. From my side for 102X everything is done, I dont know if we want to run a last test with the bool set to false, otherwise you can merge.

finnlabe:
Test were crashing due to a non-initialized vector for data with `doL1TriggerObjects == false` (did not expect any changes in that place, was wrong). Should be working for all cases now.

PR #1565, opened 2021-05-07
[106X] Updating to recommended CMSSW version and adding ParticleNet

stalbrec:
This PR:
 - updates the CMSSW version to `10_6_20`, which is the one recommended by PdmV (for all legacy analyses - see [PdmV Twiki](https://twiki.cern.ch/twiki/bin/view/CMS/PdmVRun2LegacyAnalysis))
    - for this I also cherry-picked #1550 partially (for Puppiv15 and DeepAK8-binding fix)
 - adds ParticleNet scores to `TopJet.h` and the necessary implementation in `ntuple_generator.py` and `NtupleWriterJets.cxx` in order to fill those.

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -175,6 +175,47 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
         'pfDeepBoostedJetTags:probZbb'
     ]
 
+    # from RecoBTag.ONNXRuntime.pfParticleNet_cff import _pfParticleNetJetTagsAll as pfParticleNetJetTagsAll
+    pfParticleNetJetTagsAll = [
+    'pfParticleNetJetTags:probTbcq',
+    'pfParticleNetJetTags:probTbqq',
+    'pfParticleNetJetTags:probTbc',
+    'pfParticleNetJetTags:probTbq',
+    'pfParticleNetJetTags:probTbel',
+    'pfParticleNetJetTags:probTbmu',
+    'pfParticleNetJetTags:probTbta',
+    'pfParticleNetJetTags:probWcq',
+    'pfParticleNetJetTags:probWqq',
+    'pfParticleNetJetTags:probZbb',
+    'pfParticleNetJetTags:probZcc',
+    'pfParticleNetJetTags:probZqq',
+    'pfParticleNetJetTags:probHbb',
+    'pfParticleNetJetTags:probHcc',
+    'pfParticleNetJetTags:probHqqqq',
+    'pfParticleNetJetTags:probQCDbb',
+    'pfParticleNetJetTags:probQCDcc',
+    'pfParticleNetJetTags:probQCDb',
+    'pfParticleNetJetTags:probQCDc',
+    'pfParticleNetJetTags:probQCDothers',
+    'pfParticleNetDiscriminatorsJetTags:TvsQCD',
+    'pfParticleNetDiscriminatorsJetTags:WvsQCD',
+    'pfParticleNetDiscriminatorsJetTags:ZvsQCD',
+    'pfParticleNetDiscriminatorsJetTags:ZbbvsQCD',
+    'pfParticleNetDiscriminatorsJetTags:HbbvsQCD',
+    'pfParticleNetDiscriminatorsJetTags:HccvsQCD',
+    'pfParticleNetDiscriminatorsJetTags:H4qvsQCD',
+    'pfMassDecorrelatedParticleNetJetTags:probXbb',
+    'pfMassDecorrelatedParticleNetJetTags:probXcc',
+    'pfMassDecorrelatedParticleNetJetTags:probXqq',
+    'pfMassDecorrelatedParticleNetJetTags:probQCDbb',
+    'pfMassDecorrelatedParticleNetJetTags:probQCDcc',
+    'pfMassDecorrelatedParticleNetJetTags:probQCDb',
+    'pfMassDecorrelatedParticleNetJetTags:probQCDc',
+    'pfMassDecorrelatedParticleNetJetTags:probQCDothers',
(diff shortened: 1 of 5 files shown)


core/python/ntuple_generator.py
```diff
@@ -1187,23 +1228,21 @@ def add_fatjets_subjets(process, fatjets_name, groomed_jets_name, jetcorr_label=
...
+    UpdatePuppiTuneV15(process, not useData)
 
     # Update DeepBoosted training to V2 for everything but 2016v2
     # Check https://twiki.cern.ch/twiki/bin/view/CMS/DeepAKXTagging for latest recommendations
     # e.g. 2018 specific training
     if (year != "2016v2"):
-        from RecoBTag.MXNet.pfDeepBoostedJet_cff import pfDeepBoostedJetTags, pfMassDecorrelatedDeepBoostedJetTags
-        from RecoBTag.MXNet.Parameters.V02.pfDeepBoostedJetPreprocessParams_cfi import pfDeepBoostedJetPreprocessParams as pfDeepBoostedJetPreprocessParamsV02
-        from RecoBTag.MXNet.Parameters.V02.pfMassDecorrelatedDeepBoostedJetPreprocessParams_cfi import pfMassDecorrelatedDeepBoostedJetPreprocessParams as pfMassDecorrelatedDeepBoostedJetPreprocessParamsV02
+        from RecoBTag.ONNXRuntime.pfDeepBoostedJet_cff import pfDeepBoostedJetTags, pfMassDecorrelatedDeepBoostedJetTags
```
stalbrec, line 1239: Are the differences seen by @MatthiesC (see [comment in 1550](https://github.com/UHH2/UHH2/pull/1550#issuecomment-784618411)), that are caused by this explained?

I **think** for Summer20 these options are the default? 

I can run further checks on this, in case nobody knows for certain.

stalbrec:
I forgot to handle xCone and HOTVR in terms of ParticleNET.
~I will look into this and implement it~ 
Okay it seems other tagger aren't available for them as well, which is plausible i guess.

stalbrec:
I think implementing the taggers/btags for HOTVR and xCone could/should be part of a separate PR.

stalbrec:
As discussed, I dropped the commit with the updated MINIAOD source files for now.

mschrode:
Thanks @stalbrec ! I assume the test outcome is OK and the differences are the ones expected from adding the new fields? If you can quickly confirm, then I'll merge the PR.

stalbrec:
Hi @mschrode,
I had a look at most of the summary pages. And a common difference among all of them is, that some hists for CHS jets are in the category "Diff Mean RMS". 
However the quoted numbers on the plot are identical w.r.t `ref` and `new`:

jetsAk8CHSSubstructure_SoftDropCHS.btag_DeepBoosted_HbbvsQCD

(taken from test: `https://uhh2-integration-results.web.cern.ch/uhh2-integration-results/UHH2integration/test1565/mc_UL17_TTSemiLeptonic_figs/jetsAk8CHSSubstructure_SoftDropCHS.btag_DeepBoosted_HbbvsQCD.pdf`)

I guess the difference is just smaller than the quoted precision? I can double check in the Ntuples.
Are the Ntuples that are generated for the tests kept somewhere (on EOS)? 
Otherwise I'll generate some locally.

mschrode:
Hi @stalbrec , not sure what's happening with the RMS. I updated sth in the plotting logic related to the histogram ranges, but I think this didn't change anything here (this is why I retriggered the tests).

I'll take a look at the RMS, why they are listed as different. In parallel, if you want to check on your end, the automatically produced ntuples are not stored, but they are produced with [this](https://github.com/UHH2/UHH2/blob/RunII_106X_v1/core/python/ntuplewriter_mc_UL17.py) config, taking the first 500 events from the file hardcoded therein.

Other than that, I don't quite understand, shouldn't there be new histograms appearing only for "new" (blue line) for the ParticleNet variables? I can't seem to find them in the list of plots.

stalbrec:
Hi @mschrode , for the ParticleNet variables i found the new histograms in the category "Added". There only the "new" blue histograms are filled.
(As discussed they are only filled for the "default" TopJet collections - XCone and HOTVR needs special treatment in the NTupleWriter)

I will have a look at the RMS as well. Thanks for the pointers! :)

mschrode:
Ah, my bad, I looked at the AK*4* jets, not AK*8*... - OK, great, now I see the added ones :)

stalbrec:
I managed to reproduce the the plot i posted earlier (Number of entries in the Hists match as well):
```
Branch: jetsAk8CHSSubstructure_SoftDropCHS.m_btag_DeepBoosted_HbbvsQCD
ref mean 0.06786673516035079956054687500000000000000000000000
new mean 0.06786693632602691650390625000000000000000000000000
ref RMS 1.72759783267974853515625000000000000000000000000000
new RMS 1.72759795188903808593750000000000000000000000000000
max diff among entries 0.0016263898942270316%
```

Screenshot 2021-05-12 at 10 37 00

Here one can see, that the mean and RMS are exactly the same within the accuracy quoted on the test plots, but differ at a higher decimal place.

(plot created with [this notebook](https://gist.github.com/stalbrec/bbe060e8234f6f5d911c2dbe718da63d))

mschrode:
OK, thanks a lot @stalbrec for the offline checks. I'll then merge this PR so other developments can go on top of it. (I still need to check what is happening with the RMS in the tests, but I assume this does not point to any problem with your PR.)

mschrode:
Hi @stalbrec , so there is a relative difference of 1E-6 allowed for mean/RMS to be different ([relevant method](https://gitlab.cern.ch/mschrode/UHH2-integration/-/blob/master/scripts/compareTreeDumps.py#L414)). Above that level, it is reported as different. This is consistent with your findings.

Could the difference you find be related to the different CMSSW version, eg different architecture somewhere? (It's not in anything ParticleNet related, no?)

Issue #1570, opened 2021-06-14
EGamma postreco tools


finnlabe:
Currently, we are using custom code to process electrons and photons (applying IDs, corrections...). However, the EGamma POG provides what they call "post reco tools", a centrally updated python cmssw config file that handles all this. The twiki page linked [here](https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammaPostRecoRecipes) explains this in more detail. This includes the changes needed to process samples older than 106X (for example 102X) in 106X and forward, which is not possible otherwise due to changes in egamma data structure.

I thought I would open this issue to discuss how to proceed with this: do we want to switch to this central code?
In my opinion, it is beneficial to use this centrally updated module, as this will make it very easy for us to make sure that we use everything in the way the POG suggests and are up to date.

I am looking for opinions on this. I have also started to look into the actual implementation, and will comment these technical details below.

mschrode:
Hi @finnlabe , thanks! I'm fully in favour of using centrally-provided tools as much as possible. So, if there is no strong reason *not* to use the post reco tools, I think we should switch to using them, even before we start the UL production.

finnlabe:
Find below a few lines on a possible implementation of the central code.

In principle, this is very straight-forward: the module provides a central function, which creates a single sequence that needs to be added to the path. This looks like this:

```python
# including the package
from EgammaUser.EgammaPostRecoTools.EgammaPostRecoTools import *
 
# This is the main function:
setupEgammaPostRecoSeq(process, era='2018-Prompt') # era set fixed 2018-Prompt for testing
 
# it creates a sequence called "egammaPostRecoSeq" which can be added to the path
process.p = cms.Path(
        process.content *
        process.egammaPostRecoSeq * # new sequence added here
        process.MyNtuple
    )
```

This code updates the slimmedElectrons collection. In it, IDs are set as "userInts", similar to our slimmedElectronUSER collection. However, we store keys in our Ntuples via a tag mechanism, as seen here in [ntuple_generator.py](https://github.com/UHH2/UHH2/blob/RunII_102X_v2/core/python/ntuple_generator.py#L2123) (Line 2123). These tags are not set, so simply replacing our slimmedElectronsUSER by the new slimmedElectrons using the EgammaPostRecoSeq does not work. I am not sure what the best way to proceed would be: either we continue using the tag mechanism, which probably would require a slimmedElectronsUSER class again, or find a different way for the Ntupler to get the IDs from the "userInts"? 

[...]

PR #1571, opened 2021-06-16
[106X] Replacing our Egamma code by POG postrecotools

finnlabe:
This is a PR related to [this issue](https://github.com/UHH2/UHH2/issues/1570), see there for more information. Details on the implementation can be found in [this Confluence page](https://confluence.desy.de/x/DMKXD) (I hope I put this to public, if not let me know).

This PR will surely not work perfectly, the main goal is to use the automatic tests to see which histograms show differences (or even are empty).


finnlabe:
Ah. Of course this can not work, as the postrecotool package is not installed. Is it possible to test with this aditional package @mschrode? Or is this too much hassle. I think for later deployment this would need to be part of the installation script then as well.

mschrode:
Hi @finnlabe , exactly, can you add the relevant steps to the installation scripts? These are executed for the tests as well, so in my incomplete understanding of the workflows, this should do the trick

finnlabe:
@mschrode I added the needed lines to the installation script and tested locally. Could you trigger the automatic tests again? Thanks!

mschrode:
Thanks! Let's see what happens now. Eventually, can you also update the other installation script for .csh? (Same logic, just for a different shell.) But can wait, only needed before the final merge

finnlabe:
There is one more bug, which I could have expected, sorry. I will fix this and push another change. I only implemented and tested 2018 EOY and UL, including all others should solve this.

Edit: there seems to be another issue with 2017, I will check and push a second commit once I fixed it.

finnlabe:
Ah sorry I worded that badly: I pushed a first part of the fix, but noticed another error when testing offline which I am currently trying to debug related to prefiring in 2017. So I am pretty sure these tests will fail as well.

finnlabe:
Crashed in automatic tests resulted from python errors in EOY ntuple writers for 2016 and 2017. As these do not work anyway at the moment, they are deleted to enable testing of 2018 EOY and UL for all years.

finnlabe:
There are errors in the cmsrun scripts, I will check what is going on there.

finnlabe:
This commit should fix the issue for 2017 and 2018. 2016 looks different, I need to find out what causes the error there.

Edit: the 2016 issue is the following, which I can reproduce offline:

```
edm::FileInPath unable to find file EgammaAnalysis/ElectronTools/data/ScalesSmearings/Run2016_UltraLegacy_postVFP_RunFineEtaR9Gain_scales.dat anywhere in the search path.
```

And indeed, when trying to follow the search path, I only find:

```
-rw-r--r-- 1 cvmfs cvmfs 2877561 Nov 12  2020 Legacy2016_07Aug2017_FineEtaR9_v3_ele_unc_scales.dat
-rw-r--r-- 1 cvmfs cvmfs     434 Nov 12  2020 Legacy2016_07Aug2017_FineEtaR9_v3_ele_unc_smearings.dat
-rw-r--r-- 1 cvmfs cvmfs  340524 Nov 12  2020 Run2017_17Nov2017_v1_ele_unc_scales.dat
-rw-r--r-- 1 cvmfs cvmfs     675 Nov 12  2020 Run2017_17Nov2017_v1_ele_unc_smearings.dat
-rw-r--r-- 1 cvmfs cvmfs  670192 Nov 12  2020 Run2017_24Feb2020_runEtaR9Gain_v2_scales.dat
-rw-r--r-- 1 cvmfs cvmfs     601 Nov 12  2020 Run2017_24Feb2020_runEtaR9Gain_v2_smearings.dat
-rw-r--r-- 1 cvmfs cvmfs 5495500 Nov 12  2020 Run2018_29Sep2020_RunFineEtaR9Gain_scales.dat
-rw-r--r-- 1 cvmfs cvmfs     725 Nov 12  2020 Run2018_29Sep2020_RunFineEtaR9Gain_smearings.dat
-rw-r--r-- 1 cvmfs cvmfs  648023 Nov 12  2020 Run2018_Step2Closure_CoarseEtaR9Gain_v2_scales.dat
-rw-r--r-- 1 cvmfs cvmfs     594 Nov 12  2020 Run2018_Step2Closure_CoarseEtaR9Gain_v2_smearings.dat
```

I am not sure how to continue here - does this require maybe the switch to a newer CMSSW version? In `CMSSW_10_6_26` the needed files are available. @mschrode do you have an opinion?

finnlabe:
Rebased (and squashed for less clutter), I will check if everything works offline and then restart the tests.

mschrode:
Hi @finnlabe , looks good now! Did you have any further tests/changes in mind? If not, I would merge it (I think you can even do it yourself - go ahead from my side!)

finnlabe:
Hi @mschrode, I agree that the tests look good, at least by comparing by eye. I think there are a few questions to ask before we can merge:

1) I needed to drop two IDs (`heepElectronID-HEEPV60` and `cutBasedElectronHLTPreselection-Summer16-V1`) as they were not avaiable using the postrecotools. In Mattermost, nobody said they were using those - is it then safe to drop them (I think they were used only for 2016 anyway)? If yes, maybe they should be removed from the Electron.h class to avoid confusion.

2) Two other IDs I needed to change the name: `mvaPhoID-Fall17-iso-V2-wp80` to `mvaPhoID-RunIIFall17-v2-wp80` and the same for the wp90 version. I strongly assume these are identical, but wanted to mention it here again. I made sure that from the analist side, this change is not visible: there these IDs still have the old name.

3) With this change, the electron and photon collections have a new name: `slimmedElectrons` instead of `slimmedElectronsUSER`, which would need to be changed in the xml files of all analyses using this banch. I think this is no big deal, as most are still in 102X, but its still a change worth mentioning. In case you think this is not good, one could also rename the branch to `slimmedElectronsUSER` for consistency.

4) Finally, also mainly to have this written here again: this PR drops all support for creating EOY 2016 and 2017 Ntuples with this branch. This previously already did not work if one wanted to include Egamma objects, and now all `ntuplewriter_*_2016*.py` and `ntuplewriter_*_2017*.py` are deleted.

Do you think there is any action needed on one of these items?

EDIT: Another question would be how we want to handle updating this package?

mschrode:
Hi @finnlabe , I see, thanks for summarising! From my side, I would agree to all the changes and there implications on the user analysis steps. But yes, we should probably endorse/discuss this with the others explicitly. I'll schedule a user meeting for this week/soon, and maybe we can then wait for the outcome of that discussion till the merge.

mschrode:
In the meantime, in preparation for the merge, maybe can you extend this PR so that all depending classes/files are in line? I am thinking of:
- "1." i.e. removing the removed IDs from `Electron.h` (and anywhere else)?
- "3." adjust the `example.xml` configs with the new collection names?

finnlabe:
Done. Will run tests again just to be sure, but these commits should not change anything.

finnlabe:
So this change should do the trick regarding "renaming" the slimmedElectrons collection to slimmedElectronsUSER. Renaming in this case simply means creating a copy. This change has another benefit - we can now directly compare histograms here, as the collection now is seen as the same.

I will start automatic tests, and test the running on 102X samples offline.

finnlabe:
Offline tests successful: I set up a simple analysis, just running CommonModules with an exemplary ElectronID and PhotonID and filling some histograms. No errors observed, the histograms are filled as expected.

So it seems to work for old 102X samples for all three years (2016v3, 2017v2 and 2018) and newly generated samples for 2018, UL16, UL17 and UL18.

finnlabe:
Something is strange in the tests @mschrode: do you have an Idea what is going on there?

The data jobs seem to timeout.

mschrode:
I just restarted the jobs. The file systems are rather slow lately plus I think we have a much larger file load since I added further test histos. I think I need to stream line this a bit. But it should not be an issue with your PR

mschrode:
Hi @finnlabe , tests are done! There are 11 variables with a different mean (I think the new version is not filled). Is that expected?

finnlabe:
Ah that is good to see! I will check what happens for these variables this afternoon.

finnlabe:
To comment on these variables: these are the user variables we set previous by our own slimmedElectronsUSER producer. However, some of these are available from the postrecotools already. Below I list the available ones, and then those which need to be added (if they are to be kept):

`mvaGeneralPurpose` = `ElectronMVAEstimatorRun2Spring16GeneralPurposeV1Values` -> available directly
`mvaHZZ` = `ElectronMVAEstimatorRun2Spring16HZZV1Values` -> available directly
`ElectronMVAEstimatorIso` = `ElectronMVAEstimatorRun2Fall17IsoV2Values` -> available directly
`ElectronMVAEstimatorNoIso` = `ElectronMVAEstimatorRun2Fall17NoIsoV2Values` -> available directly

These four were calculated (and renamed) by our producer - postrecotools does not rename them, so they can be fetched "directly" under their real names. This was changed in `NtupleWriterLeptons.cxx`.

This leaves two more sets of parameters. First: `effArea`, set [here](https://github.com/UHH2/UHH2/blob/RunII_106X_v1/core/python/ntuple_generator.py#L1992). I am not sure what exactly this does. I kept it around as it was previously, but we may need to check whether this is available in the postrecotools (is it not as User variable, and I am not sure whether the pat:electrons contain something similar). The same is true for the several `el_isovals`, set [here](https://github.com/UHH2/UHH2/blob/RunII_106X_v1/core/python/ntuple_generator.py#L1991). I have kept those as "custom" produced as well, as I am not sure how these are defined and how we could replace them. There are some Isolation-related things in the electron collection - maybe someone knows more regarding this?

For the moment, I will test whether, with these changes, all differences disappear.

finnlabe:
Sorry to bother you again @mschrode but we do have timeouts once more... Maybe I chose a bad time of day to start the tests :D

finnlabe:
Ok, so there are still some differences visible, I check what is going on.

Somehow one file was not pushed correctly, (and my local git was corrupted in the process). I fixed this and pushed again, starting test again, sorry for this.

finnlabe:
Now this looks good I think. Test runs on both new and old samples also still work offline, so from my side there is nothing more I would like to check.

mschrode:
Thanks! Since there was another PR #1573 merged in the meantime, I will just rerun the tests once more to make sure nothing breaks (very sure it should not interfere)

finnlabe:
Thanks @mschrode , just a question for my understanding: to test with the newly merged PR, wouldn't a rebase be needed to include those changes as the base for this PR?

mschrode:
Mh, good point, you might be right, I always get confused by this (and don't fully understand it from reading some posts). If that's not too much trouble, can you quickly rebase this RP?

finnlabe:
Sure! However, I am currently sitting in the TSG meeting, I will try but I may only manage after 16:00. When done I will restart the tests.

finnlabe:
Looks good! From my side this can be merged.

PR #1572, opened 2021-06-18
[106X] Update to CMSSW_10_6_26

finnlabe:
Switching to new CMSSW version, needed for egamma postrecotools. It compiled for me locally, and a few exemplary ntuplewriters worked, lets see what the full tests say.

### DIFF
--- a/scripts/install.csh
+++ b/scripts/install.csh
@@ -20,7 +20,7 @@ else
     echo "Please log into one and run this again"
     exit 1
 endif
-set CMSREL=CMSSW_10_6_20
+set CMSREL=CMSSW_10_6_26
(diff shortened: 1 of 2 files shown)


finnlabe:
Looks like this works! Should we then just merge, or is there anything else to check?

mschrode:
Thanks @finnlabe ! We can then rebase #1571 to this new HEAD and see if that works

finnlabe:
Great! The label for #1571 is still on "running" do the tests need to be stopped before doing that?

mschrode:
Sorry, I missed this. No, you can start new tests any time - but I saw you did this already :)

PR #1573, opened 2021-06-28
[106X] adding ParticleNet mass regression to TopJets

stalbrec:
This adds the ParticleNet mass regression as another float to the TopJets.
It was introduced in `CMSSW_10_6_25` (see [ReleaseNotes](https://cmssdt.cern.ch/SDT/ReleaseNotes/CMSSW_10/CMSSW_10_6_25.html) and PR cms-sw/cmssw#33588).

### DIFF
--- a/core/include/TopJet.h
+++ b/core/include/TopJet.h
@@ -193,6 +193,7 @@ public:
       m_btag_MassDecorrelatedParticleNetDiscriminatorsJetTags_XccvsQCD=-2;
       m_btag_MassDecorrelatedParticleNetDiscriminatorsJetTags_XqqvsQCD=-2;
 
+      m_ParticleNetMassRegressionJetTags_mass = -999.;
 
   }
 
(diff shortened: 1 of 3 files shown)


stalbrec:
The test pipeline has timeout issues in the deploy step - similar to the ones in Finn's PR - I think. 
Should we wait before restart, @mschrode ?

mschrode:
I restarted the failed steps. The problem is the run time, I need to parallelize further the jobs but didn't get the time yet to do it. For now, I have increased the allowed run time - let's see

mschrode:
Hi @stalbrec , jobs are done now. I suppose the particle net is only applicable for AK8 jets? Why does it only add a "mass" observable? Shouldn't there be more variables added?

stalbrec:
Hi @mschrode , yes to my knowledge this only applies to AK8 jets. 
I just added this one more mass observable (which was introduced in `CMSSW_10_6_25` with the PR i mentioned above) - all the other available ParticleNet observables were added in #1565 already.

mschrode:
Ah OK, thanks for reminding me, of course... I'll merge the PR

PR #1575, opened 2021-07-02
[106X] Added L1 objects for MC and data

finnlabe:
This is the same content as in #1564, now cherry-picked to 106X as the other things are done. The default behavior does not include include the L1 objects. Should we do another test with those included after the "regular" case went through?

### DIFF
--- a/core/include/Event.h
+++ b/core/include/Event.h
@@ -6,7 +6,7 @@
 namespace uhh2 {
 
 /** \brief The event container.
- * 
+ *
  * Based on the UHHanalysis' BaseCycleContainer. This class allows easy (direct) access to many frequently needed
  * objects, such as jets, electron, muons, etc. To store data beyond those directly declared data
  * members, use the GenericEvent methods, which allows settings and getting arbitrary data via handles.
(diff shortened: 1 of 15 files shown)


mschrode:
Hi @finnlabe , tests ran successfully. What I didn't fully understand: why are the L1 objects *not* included in the "max" version of the test? [This](https://github.com/UHH2/UHH2/blob/RunII_102X_v2/core/python/optionsParse.py#L45) should have included them? Is the switch called differently now?

finnlabe:
So this is indeed confusing - the switch has the correct name, as you can see [here](https://github.com/finnlabe/UHH2/blob/RunII_106X_v1_L1Info/core/python/ntuple_generator.py#L2385). And for me locally, this worked fine...

Found it! In 102X it is correctly named in optionsParse, but in 106X it was still named doL1Seed. This commit changes it, will run tests again.

mschrode:
Hi @finnlabe , looks good now, three L1seed object collections added for 'max', and from a quick check, the distributions look reasonable. Anything else to check from your side before we merge?

finnlabe:
Looks fine from my side, no more checks needed that I could think of.

PR #1576, opened 2021-07-09
[106X] added flag for photon sources and switched them on (as default)

jabuschh:
This PR fixes the empty m_source_candidates branch (with the leaves: key, px, py, pz, E) in the slimmedPhotonsUSER collection.
The boolean set to false is replaced by a flag which is switched on as default.
The file size increase for the default UL18 test sample is ~1%.

### DIFF
--- a/core/plugins/NtupleWriter.cc
+++ b/core/plugins/NtupleWriter.cc
@@ -40,81 +40,81 @@ using namespace std;
 namespace{
 
 
-template<typename T>
-void branch(TTree * tree, const char * bname, T t){
+  template<typename T>
+  void branch(TTree * tree, const char * bname, T t){
     if(tree){
-        tree->Branch(bname, t);
+      tree->Branch(bname, t);
     }
-}
-
-template<typename T>
-void branch(TTree * tree, const char * bname, const char * type, T t){
+  }
+  
+  template<typename T>
+  void branch(TTree * tree, const char * bname, const char * type, T t){
     if(tree){
-        tree->Branch(bname, type, t);
+      tree->Branch(bname, type, t);
     }
-}
+  }
 
 }
 
 namespace uhh2 {
 
 
-class CMSSWContext: public uhh2::Context {
-public:
+  class CMSSWContext: public uhh2::Context {
+  public:
     // does not take ownership of outfile
     CMSSWContext(GenericEventStructure & ges_, TFile * outfile_, TTree * outtree_): uhh2::Context(ges_), outfile(outfile_), outtree(outtree_){
     }
 
     virtual void put(const std::string & path, TH1 * h) override {
-        if(!outfile) return;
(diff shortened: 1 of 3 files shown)


mschrode:
Hi @jabuschh , thanks! Looking at the test plots, it seems nothing has changed, though. Which variables should have been added?

mschrode:
Another thing, sorry, I should have mentioned earlier: can you add the new photon switch `save_photon_keys` also [here](https://github.com/UHH2/UHH2/blob/RunII_106X_v1/core/python/optionsParse.py#L42-L45)? This control the automatic plots in the CI, giving the option to produce one set of plots with all optional plots included (`max`) and w/o them.

jabuschh:
> Hi @jabuschh , thanks! Looking at the test plots, it seems nothing has changed, though. Which variables should have been added?

I see... The variables added (leaves: `key, px, py, pz, E`) are stored in the sub-branch `m_source_candidates` in `slimmedPhotonsUSER`. It seems to me that these variables are not plotted during the tests.

> Another thing, sorry, I should have mentioned earlier: can you add the new photon switch `save_photon_keys` also [here](https://github.com/UHH2/UHH2/blob/RunII_106X_v1/core/python/optionsParse.py#L42-L45)? This control the automatic plots in the CI, giving the option to produce one set of plots with all optional plots included (`max`) and w/o them.

Sure, I will do this.

mschrode:
> > Hi @jabuschh , thanks! Looking at the test plots, it seems nothing has changed, though. Which variables should have been added?
> 
> I see... The variables added (leaves: `key, px, py, pz, E`) are stored in the sub-branch `m_source_candidates` in `slimmedPhotonsUSER`. It seems to me that these variables are not plotted during the tests.

Ah OK, makes sense

> 
> > Another thing, sorry, I should have mentioned earlier: can you add the new photon switch `save_photon_keys` also [here](https://github.com/UHH2/UHH2/blob/RunII_106X_v1/core/python/optionsParse.py#L42-L45)? This control the automatic plots in the CI, giving the option to produce one set of plots with all optional plots included (`max`) and w/o them.
> 
> Sure, I will do this.

:+1: thanks

jabuschh:
Ah yes, I just saw that those source candidate plots are available for electrons and muons: `slimmed{Electrons,Muons}USER.source_candidates.{E,key,px,py,pz}.pdf`.
I think we should add the `slimmedPhotonsUSER.source_candidates.{E,key,px,py,pz}` plots to the tests as well then?

mschrode:
> Ah yes, I just saw that those source candidate plots are available for electrons and muons: `slimmed{Electrons,Muons}USER.source_candidates.{E,key,px,py,pz}.pdf`.
> I think we should add the `slimmedPhotonsUSER.source_candidates.{E,key,px,py,pz}` plots to the tests as well then?

OK! Yes, agreed. I need to check how this is done

PR #1579, opened 2021-08-23
[106X] add L1/reco matching for leptons

jabuschh:
This PR includes the L1/reco matching for both electrons and muons.

A new field is added for each lepton collection: ```m_minDeltaRToL1Electron, m_minDeltaRToL1Muon``` which give the minimum deltaR between the reco level lepton and the matched L1 lepton. If there's no L1 object to match, the value defaults to ```+10``` (positive value such that a cut excludes these cases). See also attached screenshots below using the default TTToSemiLeptonic test sample for UL18 with 500 events.

The file size of the ntuple should not increase significantly (in my tests: <1%).
That's why I didn't include a switch, but store these two field as default.

Bildschirmfoto 2021-08-23 um 14 49 42

Bildschirmfoto 2021-08-23 um 14 50 53

### DIFF
--- a/core/include/Electron.h
+++ b/core/include/Electron.h
@@ -99,7 +99,7 @@ class Electron : public RecParticle {
     m_pfMINIIso_NH_pfwgt = 0;
     m_pfMINIIso_Ph_pfwgt = 0;
 
-    m_minDeltaRToL1Electron = 10.;
+    m_minDeltaRToL1EGamma = 10.;
 
     m_source_candidates.clear();
 
@@ -154,7 +154,7 @@ class Electron : public RecParticle {
   float pfMINIIso_NH_pfwgt() const { return m_pfMINIIso_NH_pfwgt; }
   float pfMINIIso_Ph_pfwgt() const { return m_pfMINIIso_Ph_pfwgt; }
 
-  double minDeltaRToL1Electron() const{return m_minDeltaRToL1Electron;}
+  double minDeltaRToL1EGamma() const{return m_minDeltaRToL1EGamma;}
 
   const std::vector<source_candidate>& source_candidates() const { return m_source_candidates; }
 
@@ -198,7 +198,7 @@ class Electron : public RecParticle {
   void set_pfMINIIso_NH_pfwgt(float x){ m_pfMINIIso_NH_pfwgt = x; }
   void set_pfMINIIso_Ph_pfwgt(float x){ m_pfMINIIso_Ph_pfwgt = x; }
 
-  void set_minDeltaRToL1Electron(double x){m_minDeltaRToL1Electron = x;}
+  void set_minDeltaRToL1EGamma(double x){m_minDeltaRToL1EGamma = x;}
 
   void set_source_candidates(const std::vector<source_candidate>& vsc){ m_source_candidates = vsc; }
   void add_source_candidate (const source_candidate& sc){ m_source_candidates.push_back(sc); }
@@ -299,7 +299,7 @@ class Electron : public RecParticle {
   float m_pfMINIIso_NH_pfwgt;
   float m_pfMINIIso_Ph_pfwgt;
 
-  double m_minDeltaRToL1Electron;
+  double m_minDeltaRToL1EGamma;
 
   std::vector<source_candidate> m_source_candidates;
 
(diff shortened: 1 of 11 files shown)


mschrode:
Hi @jabuschh , do you know why the matching fails (10) so much more often for muons? What dataset is this?

jabuschh:
> Hi @jabuschh , do you know why the matching fails (10) so much more often for muons? What dataset is this?

This is the default dataset of the ```ntuplewriter_mc_UL18.py```, so TTToSemiLeptonic.
So for the muons, I directly match the reco muons with the L1 muons of type ```l1t::Muon```.
For the electrons however, there are just L1 objects of type ```l1t::EGamma```. My guess is that therefore also L1 photons are matched to reco electrons, which then falsely improve the electron matching. Could this be?

jabuschh:
Yes, @finnlabe just confirmed that the L1 trigger does not distinguish between electrons and photons.

Since the following L1 objects are available, ```EGamma, Jet, Muon, EtSum, Tau```, I will also implement the L1-reco matching of the remaining objects for completeness.

finnlabe:
Just to repeat here what I wrote to you @jabuschh on mattermost: it may be worth it to first match RECO Photons to L1 Egamma objects, clean all "used" L1 Egammas from the collection, and then do the electron matching. This then should be somewhat cleaner, although still the problem with electron-photon fakes is one that stays as it is also present for offline objects. In any way this is not needed at the scope of UHH2 samples but can be done offline by just including the full L1 collections.

One more comment: maybe one could check the N(L1Egamma) and N(L1Muon) plots? I would assume high counts of Egamma objects, which could also lead to these very low dR values.

jabuschh:
Yep, makes sense, the counts for EGamma are much higher indeed:

image image

mschrode:
Hi @finnlabe @jabuschh , OK, makes perfect sense! Then I'll wait for the update before I merge

jabuschh:
I just saw that I forgot to delete a comment in the code, which is not needed...
I will add another commit, but the code itself did not change at all.

jabuschh:
Just for reference and the sake of completeness, I also want to show the deltaR distributions for the other objects here, too: 
- photons (as expected quite similar to the electrons distribution):
image

- taus:
image

- jets (here AK4PUPPI, but AK4CHS, AK8CHS & AK8PUPPI available too):
image

- MET (PUPPI MET available too, GenMET filled with default value ```+10``` since it's treated the same codewise):
this MET matching is probably not very meaningful...
image

mschrode:
Thanks @jabuschh! Any more modifications you would like to make? Otherwise, since all tests appear to have run fine and the outputs look reasonable, I would merge

jabuschh:
Not from my side. I will ask the same in Cecilia's meeting tomorrow, since they proposed/requested this, but I'm positive that this will satisfy their needs. As said, for more detailed L1 trigger studies one has to include the whole collections anyway.

finnlabe:
Just a follow up question: do you think it would make sense to cherry-pick this to 102X to keep the branches as similar as possible? I think there should not be any technical issues preventing this.

jabuschh:
Good idea, afaik this should work in 102X, too. What do you think, @mschrode?

mschrode:
Yes, I think it should work, probably totally transparently. Then again, I think some of the L1 variables and switches might have different names in 102X - @finnlabe , is this relevant here?

finnlabe:
I think there should not be any issue since you are using the initial bxvectors and not the already stored L1Particles @jabuschh. So there is no conflict I see, and even if there is I think the tests will show it without problem.

jabuschh:
Guys, I just noticed a bug...
But I already know what the problem is and how to fix it: just one line needs to be moved.
Is it possible to somehow edit this PR or is it too late now, since it's merged already?

mschrode:
Hi @jabuschh , I'm actually not entirely if/how the merged PR can be changed. There are a few suggestions how to revert the PR, but I would any ways in fact suggest to just open a new PR and comment that it is a fix of this PR in the description. Seems the cleaner way to me, also in terms of history.

jabuschh:
Alright, I did so: [PR#1580](https://github.com/UHH2/UHH2/pull/1580).

PR #1580, opened 2021-08-26
[106X] hotfix for L1-reco matching (PR#1579)

jabuschh:
This PR is a hotfix for the L1-reco matching [PR#1579](https://github.com/UHH2/UHH2/pull/1579).
It fixes the ```BadToken``` exception which appears when ```doL1TriggerObjects``` is set to ```False``` (which is the default).
I guess, it didn't appear in the tests, because all boolean switches are switched on there?!

### DIFF
--- a/core/plugins/NtupleWriter.cc
+++ b/core/plugins/NtupleWriter.cc
@@ -634,11 +634,11 @@ NtupleWriter::NtupleWriter(const edm::ParameterSet& iConfig): outfile(0), tr(0),
     prefweightdown_token = consumes<double>(edm::InputTag(prefire_source, "nonPrefiringProbDown"));
   }
   l1GtToken_ = consumes<BXVector<GlobalAlgBlk>>(iConfig.getParameter<edm::InputTag>("l1GtSrc"));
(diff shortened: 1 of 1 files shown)


mschrode:
Thanks @jabuschh ! 

Indeed this is the reason why it did not show up in the tests. Just a few days ago I [deactivated test of "default"](https://gitlab.cern.ch/mschrode/UHH2-integration/-/commit/25b266346d715856af179642281e0fbc2c9da8d4) but kept just "all" variables in the CI to speed up the process... I suppose that was not very clever... I'll revert that now.

jabuschh:
> Indeed this is the reason why it did not show up in the tests. Just a few days ago I [deactivated test of "default"](https://gitlab.cern.ch/mschrode/UHH2-integration/-/commit/25b266346d715856af179642281e0fbc2c9da8d4) but kept just "all" variables in the CI to speed up the process... I suppose that was not very clever... I'll revert that now.

I see, yeah is probably better ;)

mschrode:
Hi @jabuschh , looks like now it works. Are the control plots OK from your side? Then I'll merge

jabuschh:
Yep, looks good to me!

PR #1568, opened 2021-05-21
[106X] prefiring weights for UL17 in ntuple generator

MatthiesC:
Followed the recipe at: https://twiki.cern.ch/twiki/bin/viewauth/CMS/L1ECALPrefiringWeightRecipe#Accessing_the_UL2017_maps
No UL16 maps available yet. Needs to be addressed before UL16 ntuple production.
The above recipe requires to pull the new `L1PrefiringMaps.root` file containing the UL17 maps. This file is not yet part of CMSSW und thus needs to be downloaded manually. I implemented this in the install scripts. If you already installed the framework before this commit, do this:

`cd <your/path/to>/CMSSW_10_6_20/src/`
`git-cms-addpkg PhysicsTools/PatUtils`
`cd PhysicsTools/PatUtils/data/`
`wget --no-check-certificate https://github.com/cms-data/PhysicsTools-PatUtils/raw/master/L1PrefiringMaps.root`
`cd ../../../`
`scram b`

PR #1586, opened 2021-10-07
[106X] update prefiring weights

MatthiesC:
Updated L1 prefiring weight producer. Now handles UL and also adds the muon prefiring weights, also for EOY (in case you are not aware: there are now not just prefiring weights for ECAL 2016/2017 but also for muons in 2016/2017/2018).
Documenation: https://twiki.cern.ch/twiki/bin/viewauth/CMS/L1PrefiringWeightRecipe

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -1949,28 +1949,51 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
     ###############################################
     # TRIGGER, MET FILTERS
     #
-    # L1 prefiring, only needed for simulation in 2016/7
-    #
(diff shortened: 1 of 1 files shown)


MatthiesC:
The old PR https://github.com/UHH2/UHH2/pull/1568 is now depreciated and can be closed

MatthiesC:
Weights are filled. Here an exemplary tree for UL18 (values are very close to 1 since only muon prefiring is present in (UL)2018 data, not ECAL prefiring which would be dominant for the other years).
Command: `cmsRun ntuplewriter_mc_UL18.py maxEvents=500 outputFile=testNtuple_UL18.root wantSummary=1`

prefiringWeights_test_ul18

PR #1587, opened 2021-10-11
[106X] better request names and more commenting in crab_template.py

MatthiesC:

- Only want to compile, but don't make + test ntuples? (e.g. change `CommonModules.cxx`) Remove the @: [only  compile]

Slightly better auto-generation of crab request names; more comments in crab_template.py
Before starting the central UL production, we should still come to a conclusion whether we want to use a new group space folder for the produced samples; the new folder name then needs to be implemented in crab_template.py as well


anmalara:
Is there a reason for having default MC splitting different than from data?  I'd keep the "FileBased" option as an alternative for extremely large MC files.

MatthiesC:
Might be my own preference to use FileBased for MC, to be honest. When I produced several UL samples this spring, I noticed that most of them have number of events per file of O(upper 10,000) such that 1 file per job was a good choice (4000 events/hour is the estimate from our UHH2 Twiki production instructions). It probably makes sense to decide case by case and prior to submitting jobs to check the number of events per file in DAS

anmalara:
It should happen for a few ttbar and QCD samples. Most of them should work fine. I’d keep the size ~ constant, and only in sporadic cases change it. If the 4000 events/hour is not a good estimate anymore, then we change that.  Can you do a PR to fix this?

MatthiesC:
What exactly do you want me to change now? Remove the lines with my recommendation of using "FileBased" for MC?

PR #1588, opened 2021-10-12
[106X] fix UL16 ntuplizers by updating EGamma tools

jabuschh:
This PR updates the usage of the EGamma tools for running scales and smearings as recommended in their [Twiki](https://twiki.cern.ch/twiki/bin/view/CMS/EgammaUL2016To2018#Recipe_for_running_scales_and_sm).

With this also the automatic tests for UL16 (all combinations of pre/postVFP and data/mc) should work again, fingers crossed.

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -26,8 +26,7 @@ from RecoBTag.SecondaryVertex.trackSelection_cff import *
 from UHH2.core.muon_pfMiniIsolation_cff import *
 from UHH2.core.electron_pfMiniIsolation_cff import *
 from PhysicsTools.SelectorUtils.tools.vid_id_tools import *
-from EgammaUser.EgammaPostRecoTools.EgammaPostRecoTools import *
-
+from RecoEgamma.EgammaTools.EgammaPostRecoTools import setupEgammaPostRecoSeq
 # NOTE: all from xxx import * must go here, not inside the function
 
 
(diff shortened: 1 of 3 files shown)


jabuschh:
[This](https://github.com/cms-egamma/EgammaPostRecoTools/commit/f38439658b0784e07c3c64927185919fa6594904#diff-11a4a60b518bf24989d481468076e5d5982884626aed9faeb35b8576fcd223e1) is actually the commit in the EgammaPostRecoTools repo which renamed the files. Since this repo is cloned in the ```install.sh```, a fresh installation of UHH2 failed when ntupling the UL16 samples and thus also the automatic tests.

jabuschh:
Looks like all UL16 cmsRun jobs with the tag ```-new``` finished successfully, see link above. It is expected that the ```-ref``` jobs fail, since a fresh installation of the framework is required (due to changes in the install script).

mschrode:
Nice! This fixes the CI :)

mschrode:
@jabuschh , sorry I just realised, can you also fix the .csh install script please?

jabuschh:
Yes, good point, forgot to do that. Sure, will do.

jabuschh:
Done, also did some minor adjustments to make it more similar to the ```.sh``` install script.
The ```.csh``` install script is probably not covered by CI tests, right?
However, the git cloning should behave exactly the same.

PR #1592, opened 2021-10-13
[106X] Removing UpdatePuppiTuneV15 from ntuple_generator

stalbrec:
This is just to perform an easy cross-check for differences, between puppi v15 from Summer20 MiniAODv2 (-> `_new`) and the recomputed puppi v15 from the `UpdatePuppiTuneV15` function (-> `_ref`). 

This can be merged, if new tests pass (now with with the test samples being UL MiniAODv2 for all 4 Years.).
The difference observed for the AK4 Puppi jets can be explained with the pT cut on the jets being different for the reclustering which is done within the Update-function, while differences in the AK8 jets are reasonable within statistical uncertainties and origin from the PUPPI calculations being based on packedPFCandidates (for the Update-function) and PFCandidates (for MiniAOD).

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -1217,9 +1217,6 @@ def generate_process(year, useData=True, isDebug=False, fatjet_ptmin=120.):
     )
     task.add(process.rekeyPackedPatJetsAk8PuppiJets)
 
-    #### update PUPPI to v15
-    from CommonTools.PileupAlgos.customizePuppiTune_cff import UpdatePuppiTuneV15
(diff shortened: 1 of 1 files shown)


mschrode:
Hi @stalbrec , verstehst Du, was da beim Kompilieren schief laeuft? Kannst Du evtl. sonst mal bei dir lokal testen, was passiert,w enn Du komplett neu installierst, also das setup script ausfuehrst? Falls sich bei den eingebundenen repos was geaendert hat, findest Du das sonst lokal nicht

PR #1593, opened 2021-10-13
[106X (& 102X)] reduce ntuple generator imports to needed functions only

jabuschh:
This PR reduces the imported functions in the ```ntuple_generator.py``` to the minimum amount and aims for a sufficient way of propagating the era tag (which still needs to be implemented).

So far, this is only a test for the CI to see if it's working -> let's see.

### DIFF
--- a/core/python/ntuple_generator.py
+++ b/core/python/ntuple_generator.py
@@ -11,25 +11,6 @@ e.g.:
 
 """
 
-
-import os
-
-import FWCore.ParameterSet.Config as cms
-from Configuration.EventContent.EventContent_cff import *
-from RecoJets.Configuration.RecoPFJets_cff import *
-from RecoJets.JetProducers.fixedGridRhoProducerFastjet_cfi import *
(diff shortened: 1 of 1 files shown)


mschrode:
Hi @jabuschh , tests look good so far, all ntuple production jobs ran successfully https://gitlab.cern.ch/mschrode/UHH2-integration/-/pipelines/3121073. There was one job failing that runs lateron and collects the results, but this should be completely independent from your changes. I restarted that job, it was probably just taking too long (due to disk access or so)

jabuschh:
Hi @mschrode, alright that's good. I will check if the eras are set properly now, I'll let you know!

jabuschh:
Hi @mschrode, we can now confirm that the ```era``` tag propagation works as desired. Thanks @abenecke for checking!

Note: In principle these code changes can be applied to the pre-UL ```102X``` branch, too.

mschrode:
OK great, thanks!

PR #1596, opened 2021-10-22
[106X_v1] added LeptonMVA Top variables and output

abenecke:
I have added the Lepton MVA variable and output for electrons and muons. 
ptRatio and bTagDeepJetClosestJet rely on the closet jet. If no jet is found these are -1. 
6 weights files for the individual years and MVAs are added to common/data/YEAR (only one for 16 exist).
In addition, I added PUPPI isolation variables for electrons and muons. 
Testing with 500 events led to no significant increase in size.


mschrode:
Hi @abenecke , I just retriggered the automatic tests. They were failing because the new weight files were not properly passed to the CI, I adjusted that (problem on CI script side, not of the PR).

PR #1597, opened 2021-10-27
[106X_v1] Updating crab and ntuplewriter templates

stalbrec:
With this the ntuplewriter used in the `crab_template.py` is chosen based on the DAS string of the provided sample.

Also, after discussion with Andrea and Christopher it seemed more reasonable to update the `ntuplewriter_*_*_leadingjetConstits.py` according to my requirements, since I seem to be the only one using those and this way it will be much easier (toggle the argument `jetConstituents =True`) for people that will submit the jobs.
So i changed the default ntuplewriter that store the leading jet Constituents to save them for the 3 leading AK8 jets and additionally for all AK8 jets with pT>170GeV.

Last, but not least I added these `*leadingjetConstits.py` for the UL16 Years.

[only compile]


MatthiesC:
Thanks for this! But right now the ntuplewriter will be chosen based on the first dataset name in the list of datasets. If a user has a list of datasets from different years, your implementation will use the wrong ntuplewriter for some datasets. It would be good to decide which ntuplewriter to use per dataset individually. This can be done by setting `config.JobType.psetName` not in `crab_template.py` but in the for-loop over the datasets in the function `ByDatasets` in `CrabScript.py` (https://github.com/UHH2/UHH2/blob/09b198f9140a87a8020cfa5a3a1ab042e9f271e4/scripts/crab/CrabScript.py#L30): `self.config.JobType.psetName = ...`

anmalara:
Yes, Steffen is already implementing the same changes you suggested, with some extra checks to be a bit more flexible. Btw, the tarball is getting large… We need to update the repo too. @finnlabe @stalbrec @Ksenia or anyone who has time, we can continue the chat on MM.

anmalara:
Thanks @stalbrec. Looks great to me. One minor thing, the default storeJetConstituents should be to False, right?

As a memo for others/future, these changes make the crab template quite flexible. Now you can submit at the same time multiple years (UL only) and datasets (MINIAOD and MINIAODSIM). The script will take care to assign them to the correct output folder and tuple writer. At the same time, these changes allow the "expert" user to overwrite the default setting and store the ntuples in different folders and use different tuple_writers (all described in the comments).

stalbrec:
Hi @anmalara ,
Yes you are right, completely missed that.
I changed it, and also added another UHH2 specific crab config field for the user to provide a outLFNDirBase prefix (i.e. for private production their user-space rather than the group space) and still let multicrab decide on the year-dependent sub-dir

stalbrec:
Okay, looking at the CI-logs it seems like the DAS-query failed to to find any samples. 
(Checking locally the taball-size is still okay: `Success: tarball size 105.0 MB is below limit of 117.0 MB`)
This seemed to have worked in the past (see PR #1591) - so i guess it was a temporary problem this time?
(For that PR it is not complaining about not finding any samples: https://gitlab.cern.ch/mschrode/UHH2-integration/-/jobs/16870376#L4053)
@mschrode should i just start the test again, to see if it works now?

stalbrec:
I looked further into this and since the log prints:
```
==========
Not found /DYJetsToLL_M-50_HT-*to*_TuneCP5_PSweights_13TeV-madgraphMLM-pythia8/RunIISummer20UL16*APV*/MINIAODSIM
==========
```
it leads me to believe, that the query itself failed.
Only then this message is printed. If the samples were simply not found the `jsondict['data']` is simply an empty list:

https://github.com/UHH2/UHH2/blob/18925d2c13496c4997562ba1c92e36c7f712c8ea/scripts/crab/DasQuery.py#L13-L22

This would suggest a temporary problem with the DAS server?

mschrode:
Hi! Just quickly w/o having followed this up in too much detail: was there a DAS query present in the original version or is this an entirely new feature of this PR? Because it might be the CI does not have the necessary grid certificate / access rights to run the query

stalbrec:
No this ist not a new feature and it should not be a problem since it is (I believe) just a http request.

mschrode:
OK, good

Yes, if you believe it could just be a temporary test, just go ahead re-triggering the tests

mschrode:
Hi @stalbrec , looks like I can merge, no? Or was there sth else you wanted to check?

stalbrec:
From my side this is okay. Thanks @mschrode !

PR #1614, opened 2021-12-03
[106X_v2] updating crab_template.py to fix multicrab with newest crab3 version

stalbrec:
removing unnecessary line with doc string for UHH2 Crab ConfigSection, since this results in error in newest version of crab.

[ci skip]

### DIFF
--- a/scripts/crab/CrabScript.py
+++ b/scripts/crab/CrabScript.py
@@ -31,6 +31,7 @@ class CrabConfig:
                 self.outLFNDirBasePrefix = self.config.UHH2.outLFNDirBasePrefix
             #delete complete UHH2 specific ConfigSection, since crab will check against internal ConfigMapping for mispelled config attributes
             del self.config.UHH2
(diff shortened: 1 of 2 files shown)


stalbrec:
There seems to be another problem when using the newest crab version. I will look into this and add possible changes to this PR

stalbrec:
Okay this shoul've fixed this, but we should probably think about changing the way we get our per-dataset options to the right place in `CrabScript.py`.
The way i implemented it is pretty ugly and has now proven to break easily with crab changing ever so slightly

MatthiesC:
Okay, thanks for this. I tested your PR and multicrab's submitting, status checking etc. works again as expected. I am in favor of merging.
Right now, everything works as expected and I am not sure if changing everything now is worth the effort ("never change a running system")

mschrode:
OK I will merge now to fix the workflow for now. More fundamental changes can come later in separate PRs.

PR #1615, opened 2022-01-25
fix read_tree for events with negative weights

kdeleo:
[106X] fix for counting weighted events to handle negative weights

### DIFF
--- a/scripts/crab/readaMCatNloEntries.py
+++ b/scripts/crab/readaMCatNloEntries.py
@@ -39,7 +39,7 @@ def read_tree(rootDir, progress_bar=None):
         print e
         return numberOfweightedEntries
     w = output.splitlines()[-1]
-    if not w.replace(".","").isdigit(): return numberOfweightedEntries
+    if not w.replace(".","").replace("-","").isdigit(): return numberOfweightedEntries
     numberOfweightedEntries = float(w)
     if progress_bar: progress_bar.update(1)
     return numberOfweightedEntries


MatthiesC:
If I understand correctly, this fixes a bug that appears when the sum of event weights in a single Ntuple is negative, correct?
Could you provide the path on PNFS of such an Ntuple where this happens?

If the if-condition in L42 is fulfilled, then 0 will be the return value since `numberOfweightedEvents` has never been changed from 0 after its initialization in L29. Basically, this should happen if the output of `countNumberEvents` is not just a number but maybe an error message or similar that somehow hasn't been caught by the try/except statements a few lines before.

Aside from your actual commit, IMO it's better to inform the user about this via printed message:
```
if not w.replace(".","").replace("-","").isdigit():
  print "Problem while trying to convert countNumberEvents output to float. Affected ntuple:", rootDir
  return numberOfweightedEntries
numberOfweightedEntries = float(w)
if progress_bar: progress_bar.update(1)
return numberOfweightedEntries
```

Could you push this change as well, please?

MatthiesC:
Nevermind. I'll push it myself

kdeleo:
Hi  @MatthiesC , 
Yes right, great that you already added a warning message if the output is not a number.
In my case, it was returning 0 even if the sum of weighted events was negative, because of the minus sign. This happens in the Heavy Higgs signal samples with interference. where almost all the ntuples contain negative weights (eg /pnfs/desy.de/cms/tier2//store/group/uhh/uhh2ntuples/RunII_106X_v2/UL18/HscalarToTTTo1L1Nu2J_m365_w36p5_int_TuneCP5_13TeV-madgraph_pythia8/crab_HscalarToTTTo1L1Nu2J_m365_w36p5_int_CP5_madgraph_pythia8_Summer20UL18_v2/220118_165932/0000/Ntuple_16.root)

finnlabe:
On this, I just discussed with Komal from KIT, who encountered an issue with `w.replace(".","").replace("-","").isdigit()`: when you have many events / large weights, the string entering here might be formated using scientific notation, such as `2.82828645e+09`. This will throw the error (not able to convert) despite being a proper number. Changing to something like `w.replace(".","").replace("-","").replace("e+","").isdigit()` would fix that, but I am wondering whether there is a better way to check for this. Probably one should just implement it as

```
try:
    float(w)
except ValueError:
    print("Problem while trying to convert countNumberEvents output to float. Output is: " + w)
    print("Affected ntuple: " + rootDir)
    return numberOfweightedEntries
```

Are there any comments on this? Otherwise I can create a new PR with this change.

finnlabe:
Pinging @MatthiesC, @stalbrec @jabuschh and @mschrode here in case you'd like to comment (should you have missed my comment on this already old PR :D )

jabuschh:
adding @kdeleo.

Looks good to me!

MatthiesC:
@finnlabe Looks good! I did a quick test whether the conversion still works if the string contains scientific notation; it does:

```
In [1]: w = "2.82828645e+09"

In [2]: w.replace(".","").replace("-","").replace("e+","").isdigit()
Out[2]: True

In [3]: float(w)
Out[3]: 2828286450.0
```

KomalTauqeer:
Hi @finnlabe, regarding the last fix related to the scientifc notation, I don't see this change merged into the repository. I was merging UHH2 default branch with mine and found a conflict on this file. May be you have missed to merge it?

finnlabe:
Hi Komal, sorry, I think I proposed the fix here, but never found the time to actually make the PR - sorry about that.

I can try to do it tomorrow, but if you have it implemented already feel free to just make the PR yourself, that would be even easier :)

KomalTauqeer:
> Hi Komal, sorry, I think I proposed the fix here, but never found the time to actually make the PR - sorry about that.
> 
> I can try to do it tomorrow, but if you have it implemented already feel free to just make the PR yourself, that would be even easier :)

Hi Finn, No problem at all. I just though I should mention in case you have missed. About the PR, I am not really sure how will I do that. You can do this whenever you get time :)

PR #1620, opened 2022-02-21
[106X] Adapted b-tagging SF application module for UL 

finnlabe:
This PR changes the b-tagging SF application module to work with the updated CSV files.

Slides summarizing the changes can be found [here](https://indico.desy.de/event/33014/) - the working points and jet flavors were renamed. The new names are "converted" to the old ones to keep our code working with minimal changes.

Additionally, an issue with very large timing was found, as reported in the same meeting. Both initialization and runtime of an analysis module increased dramatically when including the module. The source was found in this code block

```
TF1 f1("", formula.c_str());  // compile formula to check validity
  if (f1.IsZombie()) {
std::cerr << "ERROR in BTagCalibration: "
          << "Invalid csv line; formula does not compile: "
          << csvLine;
throw std::exception();
  }
 ``` 
 
 as explained in some slides [here](https://indico.desy.de/event/33015/) - the TF1s do not get deleted from memory properly. As this block is only used to validate whether the CSV file is correct, if we trust that POG-provided file we do not really need this check. Therefore, I just removed it for the moment.
 
 Note: I haven't checked yet how these changes affect the WP-based corrections - maybe someone using those in UL could check? As I only changed the underlying module, I expect no problem however.
 
--
 Edit: in case there are problems with indico access, I have uploaded the slides to [this syncandshare folder](https://syncandshare.desy.de/index.php/s/4FDx5rsexTRMede).

### DIFF
--- a/common/include/BTagCalibrationStandalone.h
+++ b/common/include/BTagCalibrationStandalone.h
@@ -23,16 +23,18 @@
 class BTagEntry
 {
 public:
+  // working points and jet flavour taken from this presentation:
+  // https://indico.cern.ch/event/1096988/contributions/4615134/attachments/2346047/4000529/Nov21_btaggingSFjsons.pdf
   enum OperatingPoint {
-    OP_LOOSE=0,
-    OP_MEDIUM=1,
-    OP_TIGHT=2,
-    OP_RESHAPING=3,
+    OP_LOOSE=0, // L
+    OP_MEDIUM=1, // M
+    OP_TIGHT=2, // T
+    OP_RESHAPING=3, // shape
   };
   enum JetFlavor {
-    FLAV_B=0,
-    FLAV_C=1,
-    FLAV_UDSG=2,
+    FLAV_B=5,
+    FLAV_C=4,
+    FLAV_UDSG=0,
   };
   struct Parameters {
     OperatingPoint operatingPoint;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mschrode** [comment]
Hi @finnlabe , if you want to keep the check, is it an option to create the TF1 object with "new" and then explicitly delete at the end of the block?

**Comment by finnlabe** [comment]
Hi, I did in fact try that (using the root delete function: `f1.Delete()` at the end of the block), but for some reason this did not prevent the memory from filling up. I could not come up with another method to fix it, which is why I just deleted the lines.

**Comment by mschrode** [comment]
Ah interesting. I don't know what exactly the ROOT `Delete` method does. Did you also try the direct C++ way with 
```
TF1* f1 = new TF1(...);
do the checks...
delete f1;
```
? Maybe that works. But in general, like you say, not sure if we even need to do the check ourselves.

**Comment by mschrode** [comment]
Just for reference: using proper C++ new/delete doesn't solve the problem either. We decided to remove the check as proposed in this PR.

**Comment by finnlabe** [comment]
Just to have this here permamently: I tried the following code as suggested by @mschrode :

```
 TF1* f1 = new TF1("", formula.c_str());
  if (f1->IsZombie()) {
    std::cerr << "ERROR in BTagCalibration: "
              << "Invalid csv line; formula does not compile: "
              << csvLine;
    throw std::exception();
      }
  delete f1;
  ```
  
  Unfortunately, this still resulted in the same behavior as previously, large memory usage and long runtime.

PR #1621, opened 2022-03-08
[RunII_106X_v2] MET filter path: replace RECO path with PAT path for data

MatthiesC:
Changed the MET filter path from RECO to PAT. Only the PAT path has been updated for UL MiniAODv2, meaning that MET filters introduced in UL will be missing if using RECO. This change now includes the `Flag_BadPFMuonDzFilter` and `Flag_hfNoisyHitsFilter` in our UL MiniAODv2 ntuples.

More details: https://cms-talk.web.cern.ch/t/met-filter-badpfmuondzfilter-missing-in-ul-miniaodv2-data-samples/5101/11

Although there is no difference anymore between MC and data regarding the choice of the MET filter path (both use PAT now), I kept the `if useData ... else` structure in `ntuple_generator.py`.


mschrode:
Thanks a lot @MatthiesC for following up on this and testing&implementing the solution!

MatthiesC:
@mschrode It is not normal that the CI takes so long, is it?

mschrode:
Hi @MatthiesC , three jobs failed because they ran in the time limit, I just resubmitted them (they can fail sometimes if the diskaccess is slow). Those jobs were collecting the results, so the actual ntupler code ran fine: https://gitlab.cern.ch/mschrode/UHH2-integration/-/pipelines/3674284

PR #1627, opened 2022-04-01
[RunII_106X_v2] multicrab: fix double-slashes in ntuple paths in XML files

MatthiesC:
[ci skip]


mschrode:
Hi @MatthiesC , sorry, I missed this. Would merge it right way, but does this interfere with #1628 just opened by @anmalara ?

PR #1628, opened 2022-04-06
Run ii 106 x v2

anmalara:
Fix default path used in multicrab for creating xml files. Now taken from config file. [ci skip]


mschrode:
Hi @anmalara , does this interfere with #1627 ?

anmalara:
I answer only here. Yes, it does, but it's trivial to fix it. Once one of the two is merged, I can fix the other from the web interface.

mschrode:
👍 just merged #1627

anmalara:
commit to fix the merge conflict.

MatthiesC:
@anmalara There is a bug when you want to create multiple XML files e.g. from different years. In `crab_template.py`, the outLFNDirBase is defined as the one of the first dataset in the list:
`config.Data.outLFNDirBase = CrabYearUtilities.get_outLFNDirBase(inputDatasets[0])`
But this is in general not the correct directory for all datasets.

anmalara:
Sorry for answering only now. I see your point, but this doesn't create a conflict per se. 
I didn't think about the possibility to run multiple years at the same time. This was possible with the previous line, but the parent folder was not correctly imported in case one wanted to not use the default one. I won't be able to work on this immediately, so if someone finds an easy way to make this more general, they can do another PR.

PR #1632, opened 2022-04-19
[106X] removed defaults from JEC and JER xml file parsing

finnlabe:
[only compile]

In the recent changes by @MatthiesC to the b-tagging SF modules (see here https://github.com/UHH2/UHH2/pull/1631), JES smearing dependent handling was added. This included reading the `jecsmear_direction` parameter from the xml config file, which was deliberately set to have no default in that code, enforcing a proper setting of the variation in each config file.

I think it would make sense to have the reading of the parameters always behave in the same way, and agree with Christopher that not including a default might help in preventing mix-ups. Therefore, I removed the defaults in this PR also in the jet correction modules.

Please feel free to comment if you have any arguments for or against this change we are suggesting :)

### DIFF
--- a/common/src/JetCorrections.cxx
+++ b/common/src/JetCorrections.cxx
@@ -143,7 +143,7 @@ namespace {
 
   JetCorrectionUncertainty* corrector_uncertainty(uhh2::Context & ctx, const std::vector<std::string> & filenames, int &direction){
 
-    auto dir = ctx.get("jecsmear_direction", "nominal");
+    auto dir = ctx.get("jecsmear_direction");
     if (ctx.get("dataset_type") != "MC") {
       direction = 0;
     }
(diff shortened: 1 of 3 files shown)


mschrode:
Hi @finnlabe , did the discussion on this default conclude? Do we want to merge or are you still running tests?

MatthiesC:
From my side, this looks fine. Users will now always need to include the `jecsmear_direction` setting in their config XML when using the jet correction modules. This is in the end only a design-philosophy decision.

Quote by @anmalara  : I think there are pro and cons when having default values. Some time it's easier to have some default for non-expert users. It's the same as leaving the jes sources hard-coded. I agree with solving the inconsistency but maybe overall consistency is the way to go. Keep in mind one thing: the current code might break, so make sure you run the full test chain (and verify that the examples are still working)

From my side, we can just merge as long as the example modules all work.

finnlabe:
Thanks for posting @anmalara s comment here @MatthiesC ! I haven't found the time yet to check whether all examples that run JEC/JER corrections actually include the needed line in the config file. I'll try to do that by the end of the week.

By the way - @mschrode what checks are included in the automatic tests the bot does - are CommonModules run somewhere automatically?

mschrode:
OK thanks! I'll wait with the merge until we know the results of the checks of the examples.

The automatic tests do not touch the common modules, they just produce an ntuple and plot directly what is in there, w/o any selection or cleaning

finnlabe:
Hi everyone, I just pushed another commit that adds the required JEC and JER smearing directions to all xml files where the example module includes a call to CommonModules. Unless there is any other place I should look, or any other check needed, now the change should be complete.

PR #1633, opened 2022-04-22
fix parsing issue of btagging SF CSVs where commas in equations were …

danielhundhausen:
…not ignored


mschrode:
Hi @danielhundhausen , one pipeline of the automatic tests failed, but this was due to a time out, I just restarted it. This job was one collecting the results, all previous went fine,so I think it is safe to assume that the PR at least does not introduce any run time errors. Let's wait for the pipelines to finish so we can see the plot comparisons

MatthiesC:
I tested the PR quickly with the reshaping CSVs and it works. Can be merged.

mschrode:
Thanks @MatthiesC

PR #1635, opened 2022-05-27
include MET uncertainties for PUPPI MET

MatthiesC:
Right now, PUPPI MET uncertainties are missing in our ntuples; this PR fixes it for future productions (data members are already there; they just weren't filled...)


mschrode:
Hi @MatthiesC , thanks a lot! One small request: can you remove the if/else lines instead of commenting out? I think this is cleaner on the long run, and we have the versioning history to go back. Thanks!

MatthiesC:
> Hi @MatthiesC , thanks a lot! One small request: can you remove the if/else lines instead of commenting out? I think this is cleaner on the long run, and we have the versioning history to go back. Thanks!

Done!

PR #1637, opened 2022-05-30
b-tag SFs: custom CSV reader, consecutive adaptations for MCBTagDiscriminantReweighting, and re-work of MCBTagScaleFactor

MatthiesC:
[only compile]

The official (but no longer supported) BTagCalibrationStandalone.h is quite buggy. I noticed that it does not return the correct scale factors for the different up/down variations. I gave up on trying to debug this old code and just wrote my own CSV reader from scratch whose code length is only about 1/4 compared to the old code, which does not unnecessarily use multiple classes, ... which is just better. I put it into `namespace uhh2` out of habit but it actually is a standalone code that could be used in other environments as well.

As a consequence, I adjusted the `MCBTagDiscriminantReweighting` class (for the reshaping method) and completely re-worked the `MCBTagScaleFactor` class (for WP-based SFs). The latter one's old code, that I have now replaced, had actually used a definition of up and down variations which is not in compliance with what BTV recommends. Only one up/down variation for light and b/c flavor jets, respectively, was computed, although one should futher distinguish between components that are correlated or uncorrelated between years. Moreover, the old class symmetrized the up/down variations - something that is not recommended by BTV afaik.

I have done some tests with the new code in my setups and it now produces the correct up/down variations (all of them!) and writes them to the AnalysisTree. Some more documentation can be found in the code. Feedback from fellow b-taggers is welcome!

Bonus: You can know set custom abseta/pt bins for `BTagMCEfficiencyHists` (constructor parameters). The default bins are lined up with the bins currently available in the Run II UL CSV files.

### DIFF
--- a/common/include/JetHists.h
+++ b/common/include/JetHists.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <boost/optional.hpp>
+
 #include "UHH2/core/include/Hists.h"
 #include "UHH2/core/include/Event.h"
 #include "UHH2/core/include/LorentzVector.h"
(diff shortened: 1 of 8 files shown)


mschrode:
Hi @MatthiesC , I think the compilation error was related to the missing const qualifiers added in #1636 ? I just restarted the test after #1636 was merged

finnlabe:
Hi @MatthiesC I tested these changes in my setup, and observe large runtime issues, similar to what I fixed in https://github.com/UHH2/UHH2/pull/1620

Do you see something similar? For me, a ttbar sample after my preselection has over 36 hour of runtime on condor (usually taking much below an hour otherwise).

MatthiesC:
> Hi @MatthiesC I tested these changes in my setup, and observe large runtime issues, similar to what I fixed in #1620
> 
> Do you see something similar? For me, a ttbar sample after my preselection has over 36 hour of runtime on condor (usually taking much below an hour otherwise).

I cannot confirm this. My jobs run smoothly locally and on Condor. Do you know where the lag starts? At the initialization of the module (I added some print-outs, so this should be easily detectable) or during the event loop?

MatthiesC:
> Hi @MatthiesC , I think the compilation error was related to the missing const qualifiers added in #1636 ? I just restarted the test after #1636 was merged

Yes, my bad.

finnlabe:
The initialization is decently fast, similar to the central module. The large run-time is really during the event loop, in the same way as I observed previously. I did not look into your code in detail, but for the central module this was caused by a creating of TF1s which where then not properly deleted again afterwards -> this led to a memory overflow when running large samples on condor. For my test, everything runs fine on smaller samples, but for larger ones it jump to this high runtime, supporting the potential reason in a memory overflow.

A similar thing could be happening [here](https://github.com/UHH2/UHH2/blob/RunII_106X_v2/common/src/UHH2BTagCalibReader.cxx#L99), but this is just a wild guess.

MatthiesC:
I don't think that there is an issue at https://github.com/UHH2/UHH2/blob/RunII_106X_v2/common/src/UHH2BTagCalibReader.cxx#L99

Could you try adding `delete correct_entry;` just before `return result;` at the very end?

finnlabe:
I have rolled back to the previous UHH2 version for the moment, but when I have finished my other studies, I'll try to check that later today.

MatthiesC:
It seems that this is not related to our code: https://sft.its.cern.ch/jira/browse/ROOT-10191

There really was/is (depending on our ROOT version) a bug in `TF1`.

MatthiesC:
Indeed, we use `ROOT 6.14/09` (tested with `root -b` after setting up the `CMSSW_10_6_28` environment). This ROOT version is from 2018. The PR for the `TF1` bug fix is from 2019. I will dive into this a bit more.

finnlabe:
Very interesting find, that explains why I could not fix it for the central module without just deleting any call to `new TF1` :D

PR #1639, opened 2022-06-01
Avoid memory leak in b tag scale factor application caused by bug in TF1

MatthiesC:
[only compile]

In older ROOT versions, including version `6.14/09` that is used for `CMSSW_10_6_28` which this UHH2 branch is based on, the class `TF1` features a memory leak. This bug is fixed in newer ROOT versions. See [1].

The memory leak was avoided somehow by the changes to `BTagCalibrationStandalone` done in #1620 but was re-introduced with the new CSV reader in #1637.

Since this bug is related solely to `TF1`, we can just avoid using this class and use `TFormula` instead. This PR makes this change. Aside from the bug, it probably also makes more sense to rely on `TFormula` (a simple function interpreter) instead of `TF1` (rather a class to draw functions; not needed at all here!).

Personally, I never experienced the memory leak (for unknown reasons a.k.a. luck). Therefore, it makes sense for @finnlabe, who did, to check whether it is working now and, thus, whether the memory leak that he experienced is really caused by the memory leak in `TF1`.

[1] https://sft.its.cern.ch/jira/browse/ROOT-10191

Just a personal comment on the ROOT version that we use: I have no clue why `CMSSW_10_6_28`, that is designed for Ultra Legacy, still relies on a ROOT release from 2018...


finnlabe:
Thanks for updating that @MatthiesC I'll check the change and will update here once I have results.

mschrode:
> 
Thanks @finnlabe ! I'll wait with merging until you have run the tests (and for the automatic tests, of course)

finnlabe:
Hi @MatthiesC unfortunately this change does not solve the issue I am having, I still see the incredibly high run times for larger samples, unchanged to the situation with TF1.

MatthiesC:
This is unfortunate. I will try something else... Please don't merge/close yet.

MatthiesC:
I changed `CalibEntry.func` such that it is now a `TFormula`, not `TFormula*`. To be honest, I don't know why this should help, but this now emulates the `TmpEntry` struct in `BTagCalibrationStandalone` (https://github.com/UHH2/UHH2/blob/RunII_106X_v2/common/src/BTagCalibrationStandalone.cc#L437-L445), aside from the change to `TFormula`. Let's try this first. If it does not work, let's also change it back to `TF1`.

Btw, I still think that the problematic memory leak arises within `TF1`/`TFormula`. Despite that, it would have been good pratice of mine to write a descructor for `BTagCalib::Reader` that `delete`s all `.func` pointers initialized with `new ...` to avoid another possible (but not really problematic) memory leak. After the latest commit, this is of course not needed anymore.

finnlabe:
To reflect here what I already said in today's meeting: unfortunately also by this way of constructing the function, the issue remains.

I might double-check later today whether everything else is fine in my application, but I don't expect any error there.

MatthiesC:
See this ROOT forum entry: https://root-forum.cern.ch/t/tf1-and-memory-leak/37565

The problem described there is basically the same as the one we observe.

The answers say that there is another memory leak in `TFormula::HandleParamRanges` that is fixed in newer versions of ROOT but not ours.

[to be continued]

MatthiesC:
@finnlabe Did you actually also try `delete correct_entry;` at the very end?

finnlabe:
No, I haven't tried that yet, I first wanted to test your PRs. I just have some jobs running (which I don't want to interrupt be compiling UHH2 :D ) but I'll test that change afterwards.

MatthiesC:
New commit! As far as I understand it, the newest commit now makes all `TFormula`-related stuff work on the stack memory and avoids the heap memory. Maybe that circumvents leakages.

finnlabe:
Hi @MatthiesC, to keep you updated: I tested your new commit, which does not change the runtime. Additionally, I tested using `delete correct_entry;` at the end (without the newest commit), which crashes with the following error message:

`*** Error in "sframe_main": free(): invalid pointer: 0x000000000a7de708 ***`

(Followed by a large stack trace which I'll not past here, unless someone wants to see it).

MatthiesC:
Again, very unfortunate! I will dive more...

The memory leackage in `TFormula::HandleParamRanges`, that I mentioned earlier, is caused by an invocation of `new Ssiz_t()` without deleting it afterwards. `Ssiz_t` is just a typedef of `int`. As far as I understand, `TFormula::HandleParamRanges` is called for each `TFormula::Eval` call (but I could be wrong). If so, then the memory allocated is `sizeof(int) * O(n_events * n_jets_per_event * n_btagging_weights)` which is something like `O(1) GB` for `O(n_events) = 100,000`.

MatthiesC:
I now implemented a hacked version of `TFormula`, called `UHH2TFormula`, that has a fix for the memory leak mentioned in [1] and [2]. I initially wanted to just write a child class of `TFormula` like `class UHH2TFormula: public TFormula ...` and override the method that has the leak; but this is not an option since the problematic method is private.

I ran some tests and noticed that some of my jobs actually run shorter with the latest commit and consume less memory (checked with `condor_history <jobid> -format "%d\n" MemoryUsage`). Fingers crossed!

[1] https://sft.its.cern.ch/jira/browse/ROOT-10089
[2] https://root-forum.cern.ch/t/tf1-and-memory-leak/37565

finnlabe:
Now with this change I really would have expected this to be fixed now, but unfortunately the issue seems to be located somewhere else, I still have about 40 hours of runtime for my ttbar sample.

mschrode:
@finnlabe , do you have an example I could test in a clean install to check if the effect is somehow connected to some local settings from either of you?

finnlabe:
Well you could just pull my analysis repository from [here](https://github.com/finnlabe/TstarTstar) and run the UL18 selection step. You might need, in addition to the usual setup, to download and compile the HOTVR repo.  I am not 100% sure whether this is everything, or whether there are some other files one would need to manually add however.

Is this what you mean @mschrode  ? I can also construct a smaller example module, but that of course might change also the source of the issue, wherever it may be.

I have prepared a config file that should run my selection on ttbar from my preselection, you would only need to rename to `.xml` (GitHub does not allow uploading xml files) and replace the output directory: [FinnsSelectionTest.txt](https://github.com/UHH2/UHH2/files/8831383/FinnsSelectionTest.txt)

mschrode:
ah yes, I meant something like this. Will try

MatthiesC:
I tested @finnlabe's setup as well and the jobs run with usual runtime locally but have the incredibly long runtime on Condor. This is consistent with what @finnlabe observes.

MatthiesC:
The memory leak that is fixed now in `TFormula` cannot be the cause of the extreme lag of the jobs. The size of the leaking memory would increase for more and more events that are processed. But the lag occurs basically instantly after the SFrame jobs start processing the very first events. Also, the memory usage is not that huge in the beginning and cannot be the cause of the lag, as far as I see it.

finnlabe:
Interesting. So it must be something else in the implementation of the b-tagging SFs that leads to this then, as I can at least say with certainty that the runtime is much lower when rolling back to the previous module.

MatthiesC:
Did some checks with help of `std::chrono` print-outs in your `TstarTstarSelectionModule`. Ran it locally and on interactive condor session (`condor_submit -i`). The lag also occurs when using interative session and the runtime increase is happening all over the place in `TstarTstarSelectionModule::process`, not just at the point where you call `MCBTagDiscriminantReweighting::process` (whose _relative_ runtime compared to other code within `TstarTstarSelectionModule::process` stays the same).

MatthiesC:
**New commit that should finally fix it.** Let's summarize:

@finnlabe Observed a huge lag in his SFrame jobs, pinning it down to the module that reads the b tagging scale factors, in particular to some unidentified issue with the `TF1` class. However, this lag, that starts with the beginning of the event loop and increases the job runtime from some minutes up to 40 hours or more per job, only appears when running on the batch system, not locally. Doing some checks with `std::chrono` also showed that the lag is not introduced just at the point in the code where the scale factors are calculated for each event but all parts of the code (every selection step, calibration step, histogram fill etc.) run slower. This was accompanied by slowly but steadily increasing memory, hinting that there might be a memory leak that somehow only becomes an issue on the batch system. Nonetheless this doesn't really explain the problem since the jobs are laggy immediately after entering the event loop, when the memory usage is still low.

I did some research regarding potential memory leaks in `TF1` and found that the ROOT version that we use (fixed to `6.14/09` for `CMSSW_10_6_28`) indeed features a by-now fixed memory leak [1]. I thus just switched out `TF1` with the more basic `TFormula` class which `TF1` is based on but still provides the same needed functionality (compiling formulas given as strings to valid C functions during runtime), avoiding the `TF1` memory leak. It turned out that this did not fix it. After some more digging, it turned out that also `TFormula` itself features in that ROOT version a memory leak in its `TFormula::HandleParamRanges` method [2]. So I hacked `TFormula` and called the class `UHH2TFormula`. This indeed reduced a bit the memory usage of my jobs and their runtime by some minutes. So it was a good change to the code after all! But it did not fix the immense lag that @finnlabe observed.

At this point I have to amend my statement that I have not experienced the same issue with my analysis code as @finnlabe has. I indeed see the same strange lag, but less extensively. I tested whether there is any difference between the WP-based and the re-shaping method. And yes, there is! The only real difference between these two options (aside from the physics) is that in the WP-based case less instances of (`UHH2`)`TFormula` are loaded into memory since the number of calibration entries is less compared to the re-shaping method (O(100) vs. O(1,000) entries). The lag seems to be more heavy the more _different_ `TFormula`s are loaded. Now comes the tricky part: _different_ `TFormula`s! I tested what happens if I just set the formula for each `CalibEntry` to a fixed expression, e.g. `TFormula("", "42*x")`: the lag is gone. Now you might think that the fact that all `TFormula`s are equal objects cannot change the picture since we still have O(100-1000) individual `TFormula` objects residing in memory, but it does for some wild reason. Here one could think that maybe, just maybe, there is something wonky with the formulas provided in the CSV files which are of course all different. But no: I also tested what happens if the `TFormula`s of all O(1,000) `CalibEntry`s are different without relying on the CSV files: I just set each `TFormula` expressions to a unique constant value, like `TFormula("", "i")` with `i = 0 ... O(1000)`. The lag appeared again!

Here I stopped trying to figure out the issue. In summary: the existence of huge amounts of _different_ `TFormula` objects in memory is the reason for the lag, but it's not related to having too little memory as far as I see it. But why the issue only appears on the batch system but not locally is also beyond my understanding. I think that there is some more fundamental problem on ROOT's side, might it be somewhere in Cling (the runtime C++ interpreter, which is also used by `TFormula`) or somewhere else. I would start making random guesses at this point.

### tl;dr:
In conclusion, I decided now to move away from a ROOT-based implementation. I found a nice library that does in principle the same as `TFormula`: take a string, compile it to a valid function during runtime, and return values when giving arguments. The library is called `exprtk` [3, 4]. Everything needed is in the 40k-line `common/include/exprtk.hpp` header. Not only the lag during the event loop is gone, but also the annoyingly long initialization time of the calibration hash map at the beginning of each job is reduced to a Planck time. I of course checked whether the calculated scale factors are still the same - and they are, for both WP and reshaping! Nonetheless, I encourage @finnlabe to check this, too.

If you have concerns regarding adding a 40k-line source file to UHH2, please raise your voice. In terms of compilation time, there is no problem as far as I have measured it.

Something else to keep in mind: The string formulas given in the CSV files represent valid C code, using basic math operations (`+`, `-`, `*`, `/`, `pow()`, `exp()`) which can be interpreted by `exprtk`. If this format will ever change and e.g. will use ROOT's math functions (`TMath::Power()` etc.), then we would need to adjust the `BTagCalib::Reader::ReadCSV` method a bit to do more pre-processing of the string before passing it to `exprtk`.

[1] https://sft.its.cern.ch/jira/browse/ROOT-10191
[2] https://sft.its.cern.ch/jira/browse/ROOT-10089
[3] https://www.partow.net/programming/exprtk/index.html
[4] https://github.com/ArashPartow/exprtk

mschrode:
Thanks @MatthiesC for investigating into this - the results are unbelievable! What does ROOT do??! If @finnlabe can confirm that changing the formula interpreter does the trick, I'm all up for doing it like that.
Just one other question: @anmalara mentioned it might be possible to use the official BTV code based on the JSON inputs also with miniAOD/C++, using correctionlib. Can you confirm this is not an option for us? Otherwise, it might make sense to try that instead?

MatthiesC:
@mschrode After quickly looking at the `correctionlib` docs, it seems like we can use it - and should in principle. However, this would take some time to implement and I cannot spent time on this right now. `correctionlib` also supports interpreting formulas given as strings [1]; they emulate the basics of `TFormula` [2] with the help of the `peglib` library [3] which does stuff similar to `exprtk`. - Definitely interesting to see that they do not rely on `TFormula` itself for this, what I would have expected.

I would prefer merging this PR (if @finnlabe's checks are okay) right now and leave it open for future delevopments to switch to `correctionlib`.

[1] https://github.com/cms-nanoAOD/correctionlib/blob/master/src/formula_ast.cc
[2] This is actually mentioned in `correctionlib`'s README: https://github.com/cms-nanoAOD/correctionlib/blob/master/README.md
[3] https://github.com/yhirose/cpp-peglib

finnlabe:
This is super interesting, thanks @MatthiesC ! I'll check right away whether this fixes the issue also on my side and will update here once I have results.

Edit: I can already confirm much better looking runtimes with the new setup. Will update once the jobs are done.

finnlabe (approved):
I checked the code, to me everything looks fine. Additionally, I ran my selection module with these changes and was able to obtain results within a reasonable timescale. My tests also confirmed no changes to the SF themselves, so from my side we can merge this.

mschrode:
> @mschrode After quickly looking at the `correctionlib` docs, it seems like we can use it - and should in principle. However, this would take some time to implement and I cannot spent time on this right now. `correctionlib` also supports interpreting formulas given as strings [1]; they emulate the basics of `TFormula` [2] with the help of the `peglib` library [3] which does stuff similar to `exprtk`. - Definitely interesting to see that they do not rely on `TFormula` itself for this, what I would have expected.
> 
> I would prefer merging this PR (if @finnlabe's checks are okay) right now and leave it open for future delevopments to switch to `correctionlib`.
> 
> [1] https://github.com/cms-nanoAOD/correctionlib/blob/master/src/formula_ast.cc [2] This is actually mentioned in `correctionlib`'s README: https://github.com/cms-nanoAOD/correctionlib/blob/master/README.md [3] https://github.com/yhirose/cpp-peglib

Ah, great, thanks for checking @MatthiesC ! I agree, let's then go ahead for now with your new module, and we consider moving to correctionlib at a later time. Possibly this is also not needed any more depending on how the Run 2 analyses progress.

mschrode:
> I checked the code, to me everything looks fine. Additionally, I ran my selection module with these changes and was able to obtain results within a reasonable timescale. My tests also confirmed no changes to the SF themselves, so from my side we can merge this.

Thanks for testing @finnlabe !

mschrode:
I will then merge this PR given that all tests indicate it solves the run time issues like this. We will keep the correctionlib option as potential refinement for a later iteration. As for this, I opened issue #1640

PR #1643, opened 2022-07-02
Fix indexing of genInfo QCD weights for DY samples

danielhundhausen:
- Fixes the indices that are used to access the QCD weights (mur/muf) in the `MCWeights` module for DY samples
- After a review of the LHE record of the DY samples it was found that their weight array differs from the one used in e.g. TT
- Fixes a downward pull of up *and* down variations in DY samples for mur/f variations
- The indexing of mur/muf was filliped when compared to the LHE record of the TTTo2L2nu sample, which was fixed

### DIFF
--- a/common/include/MCWeight.h
+++ b/common/include/MCWeight.h
@@ -77,6 +77,7 @@ class MCScaleVariation: public uhh2::AnalysisModule {
 
   private:
   double syst_weight;
+  bool is_dy;
   int i_mu_r = 0, i_mu_f = 0;
   uhh2::Event::Handle<float> h_murmuf_weight_upup_;
   uhh2::Event::Handle<float> h_murmuf_weight_upnone_;
--- a/common/include/MCWeight.h
+++ b/common/include/MCWeight.h
@@ -78,13 +78,40 @@ class MCScaleVariation: public uhh2::AnalysisModule {
   private:
   double syst_weight;
   bool is_dy;
+  bool is_wjets;
+  bool is_alps;
   int i_mu_r = 0, i_mu_f = 0;
+  void initialise_handles(uhh2::Event & event);
   uhh2::Event::Handle<float> h_murmuf_weight_upup_;
   uhh2::Event::Handle<float> h_murmuf_weight_upnone_;
   uhh2::Event::Handle<float> h_murmuf_weight_noneup_;
   uhh2::Event::Handle<float> h_murmuf_weight_downdown_;
   uhh2::Event::Handle<float> h_murmuf_weight_downnone_;
   uhh2::Event::Handle<float> h_murmuf_weight_nonedown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn1_weight_upup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn1_weight_upnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn1_weight_noneup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn1_weight_downdown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn1_weight_downnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn1_weight_nonedown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn2_weight_upup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn2_weight_upnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn2_weight_noneup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn2_weight_downdown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn2_weight_downnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn2_weight_nonedown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn3_weight_upup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn3_weight_upnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn3_weight_noneup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn3_weight_downdown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn3_weight_downnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn3_weight_nonedown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn4_weight_upup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn4_weight_upnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn4_weight_noneup_;
+  uhh2::Event::Handle<float> h_murmuf_dyn4_weight_downdown_;
+  uhh2::Event::Handle<float> h_murmuf_dyn4_weight_downnone_;
+  uhh2::Event::Handle<float> h_murmuf_dyn4_weight_nonedown_;
 };
 
 
--- a/common/src/MCWeight.cxx
+++ b/common/src/MCWeight.cxx
@@ -209,13 +209,43 @@ MCScaleVariation::MCScaleVariation(Context & ctx){
   h_murmuf_weight_downnone_ = ctx.declare_event_output<float>("weight_murmuf_downnone");
   h_murmuf_weight_nonedown_ = ctx.declare_event_output<float>("weight_murmuf_nonedown");
 
+  h_murmuf_dyn1_weight_upup_     = ctx.declare_event_output<float>("weight_murmuf_dyn1_upup");
+  h_murmuf_dyn1_weight_upnone_   = ctx.declare_event_output<float>("weight_murmuf_dyn1_upnone");
+  h_murmuf_dyn1_weight_noneup_   = ctx.declare_event_output<float>("weight_murmuf_dyn1_noneup");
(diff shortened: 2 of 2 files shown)


MatthiesC:
@danielhundhausen Thanks for this!

Could you please check `WJetsToLNu` as well? It uses the same generator (madgraphMLM) as `DYJetsToLL`. I looked into my older MiniAODv1 templates for the scale variations and the issue that all variations are downwards is present for both these processes, so it is likely that for MiniAODv2 we have the same.

Aside from this, one should check whether this issue is also present for `WJetsToQQ`, `ZJetsToQQ`, `ZJetsToNuNu`. They all are produced with madgraphMLM as well. So, in the end, the boolean `is_dy`, that you defined, should likely not only check that `dataset_version` starts with `DYJetsToLL` but with `DYJets`, `ZJets`, or `WJets`. I guess it would also be good to implement a warning `cout` in the constructor of `MCScaleVariation` that different indices are used than usual.

MatthiesC:
I don't have the proper up/down templates yet, but I looked quickly into the distributions of the weights in the TBrowser in my AnalysisTrees: All mur/muf weights for `ST_s-channel` are also below 1, hinting that there will also only be a downwards shift. `ST_s-channel` is also produced with MadGraph, more precisely `amcatnlo`, i.e. MadGraph at NLO.

Then we also have the `QCD_HT` samples that are produced with madgraph(MLM). One should check if the problem arises there as well.

MatthiesC:
@danielhundhausen Could you also provide a plot of the distributions of up/down variations here to highlight that the fix works, please?

danielhundhausen:
Will continue work on this tomorrow, but for now you can find the updated variation plots [here](https://dhundhausen.web.cern.ch/AZHToTTLL/Uncertainty_Ratios/CMS_murmuf/DYJets/).

MatthiesC:
By the way: If your PR only includes changes that do not affect the ntuple generator, it is enough to run the tests with [only compile] or, if you neither want to test the compilation of UHH2, just [ci skip].

mschrode:
Hi @danielhundhausen , great you found the reason for this and implemented the fix!

Next to WJets, there is also a DYtoQQ sample that likely has the same weight table; probably it is the same for all `madgraph` samples? So maybe it is more adequate to have instead of `is_dy` the bool `is_madgraph_sample` or `has_madgraph_weight_table` or similar and then search for `madgraph` in the sample name?

MatthiesC:
@mschrode 

> Next to WJets, there is also a DYtoQQ sample that likely has the same weight table; probably it is the same for all `madgraph` samples? So maybe it is more adequate to have instead of `is_dy` the bool `is_madgraph_sample` or `has_madgraph_weight_table` or similar and then search for `madgraph` in the sample name?

According to Matteo (see mattermost), at least the TTZ samples, which are also based on amcatnlo, follow the standard convention for the weights numbering. So it is not that simple. We need to check all samples individually.

mschrode:
> @mschrode
> 
> > Next to WJets, there is also a DYtoQQ sample that likely has the same weight table; probably it is the same for all `madgraph` samples? So maybe it is more adequate to have instead of `is_dy` the bool `is_madgraph_sample` or `has_madgraph_weight_table` or similar and then search for `madgraph` in the sample name?
> 
> According to Matteo (see mattermost), at least the TTZ samples, which are also based on amcatnlo, follow the standard convention for the weights numbering. So it is not that simple. We need to check all samples individually.

Yes, but I think it is LO Madgraph (`madgraph-mlm` in the PD name) vs NLO Madgraph (`amc`) in the sample name.
Alternatively, it could also be an option to have this configurable. Eg. implement a default weight scheme, and then allow to specify in the config the relevant weights for a given sample name (or substring). This way, one can easily adjust the behaviour later should it arise for additional samples

MatthiesC:
@mschrode Doing it via a config is a good idea! This is the way. It could be implemented by having one SFrame XML config parameter that contains all relevant string matches, split by whitespaces, e.g. `'<Item Name="find_a_good_name_for_this" Value="DYJetsToLL WJetsToLNu ..."/>`

bonanomi:
> @mschrode
> 
> > Next to WJets, there is also a DYtoQQ sample that likely has the same weight table; probably it is the same for all `madgraph` samples? So maybe it is more adequate to have instead of `is_dy` the bool `is_madgraph_sample` or `has_madgraph_weight_table` or similar and then search for `madgraph` in the sample name?
> 
> According to Matteo (see mattermost), at least the TTZ samples, which are also based on amcatnlo, follow the standard convention for the weights numbering. So it is not that simple. We need to check all samples individually.

I agree with Matthias' proposal. It is true that for TTZ the plots look consistent with the expectation, but if this issue is shared among all the `amcatnlo` samples, than there is the risk that for TTZ things look fine by chance. It is probably better to cross check explicitly.

danielhundhausen:
@MatthiesC @mschrode @bonanomi 
I think we are all in agreement that the robust solution is to check all `amcatnlo` and `mlm` LHE records, to see if the indexing is correct. I will get to work on this, but could take me a week or so.

@MatthiesC @mschrode 
The advantages of having the indexing configurable in the xml are unclear to me. This seems to me like something where there is only one correct way of doing it (i.e. one correct set of indices per sample), so why would one want the overhead of a config parameter?
I propose we write several `std::map`s in the `MCScaleVariations` module (keys of the format `murUp_mufNone`, values are the indices) and automatically apply the correct one per sample.

bonanomi:
Given how this is scaling, I wonder if maybe the best solution isn't to think of a check based on `event.genInfo->systweights()` length rather than on the name of the sample.

@danielhundhausen  when you compared DY and TT, for TT were there only 8 `event.genInfo->systweights()`? If so, maybe this could be a good discriminant to decide whether we should use the "old" numbering we had implemented or the new one.

danielhundhausen:
@bonanomi 
I think this would be a decent "rule of thumb" solution. I updated my [question in CMS talk](https://cms-talk.web.cern.ch/t/qcd-scale-variation-weight-issue-with-dy-miniaod-samples/12347) with our current findings. Ideally, someone can give us a general answer without us having to check each sample.

The idea just came to me to look at the way the weights are stored in the `NanoAOD` objects. If they already handle this, a general solution could be implemented from looking at this. Does someone know where the relevant code for this can be found?

bonanomi:
Hi @danielhundhausen, I found [this](https://github.com/cms-sw/cmssw/blob/24de28d0548f0f50cfc332c6d75d0ef30a79417a/PhysicsTools/NanoAOD/plugins/GenWeightsTableProducer.cc#L660-680) for the NanoAOD case, but i'm not sure this code would help us, as after a quick look I did not find any difference between DY and other processes :/

jabuschh:
Hi, I just wanted to point out that in the official [genproductions repo](https://github.com/cms-sw/genproductions) one can actually see the different handling for the scale and PDF weights depending on LO/NLO:
- MadGraph LO (madevent): [link](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L194)
- MadGraph NLO (aMCatNLO): [link](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh#L135)

danielhundhausen:
@MatthiesC @mschrode @anmalara I propose we merge this PR as it stands, as this fixes the immediate issue multiple analyses are having right now with the DY sample. I can work on a more general solution soon but am short on time right now. I have created an issue #1644 that we can refer to.

bonanomi:
> @MatthiesC @mschrode @anmalara I propose we merge this PR as it stands, as this fixes the immediate issue multiple analyses are having right now with the DY sample. I can work on a more general solution soon but am short on time right now. I have created an issue #1644 that we can refer to.

I tend to agree. Not sure whether this was discussed offline, but Daniel in his studies figured that `event.genInfo->systweights()` contains more weights than the ones we need to evaluate QCD scales. Hence, the solution proposed above suggesting to use the length of this vector seems not to be the way to go.

@danielhundhausen can comment more on this probably.

jabuschh:
> @MatthiesC @mschrode @anmalara I propose we merge this PR as it stands, as this fixes the immediate issue multiple analyses are having right now with the DY sample. I can work on a more general solution soon but am short on time right now. I have created an issue #1644 that we can refer to.

Would also be fine with me to merge it like this for now.
@danielhundhausen Maybe you could just change the name of the bool from `is_dy` to `is_MadGraphLO` (or something similar)? Since it not only affects DY, but also my ALP signal samples (and possibly more).

anmalara:
Hi, thanks for fixing this. I don't mind this change, as long as it's flexible enough. At the moment, one cannot choose freely this parameter from the XML file.

I'm not sure if this was already discussed in this chat or on MM, but one solution could be to read one new parameter (say "is_MadGraphLO") from the XML, which is then a string of lists (separated by spaces). This string is then interpreted and compared to the `dataset_version`. Still not ideal, but rather flexible I believe. Thoughts?

mschrode:
Thanks all! Our conclusion discussed here locally was that we merge the PR now as is (basically hard-coded for specific samples), and then take more time to discuss more robust and configurable long-term solutions. These could also be to define the to-be-used weight scheme in the sample database entries for each samples, which has the charme that one has to take care of this when adding a new sample.

mschrode:
we can discuss further the long-term solutions in #1644

danielhundhausen:
@anmalara Thank you for the comment. I'm in principle open to any solution, just a little short on time right now, which is why I would like for this preliminary fix to be merged first and as @mschrode said we can find a solid solution in #1644 . Are you okay with that?
I also added the automatic writing of the dynamical scales to the output tree in the case of the "alternate indexing samples" to this PR.
I don't think the online tests ran again, but I tested it locally without issues.

PR #1647, opened 2022-08-26
[106X] add QCD flag for MC scale variations

jabuschh:
When checking the MC scale variations for all SM backgrounds used in our analysis, I noticed that the ones for QCD also were below 1 (as observed before for e.g. DY and WJets). 
We use the ```QCD_HT*``` ones. Although only some of them have the ```madgraphMLM``` tag explicitly in the sample name (somehow most UL16 samples only), this should be the case for all years.
I guess that it's the same for the pT-binned samples, that's why I added the flag for all QCD samples here.

See also https://github.com/UHH2/UHH2/pull/1643.

[only compile]

### DIFF
--- a/common/include/MCWeight.h
+++ b/common/include/MCWeight.h
@@ -79,6 +79,7 @@ class MCScaleVariation: public uhh2::AnalysisModule {
   double syst_weight;
   bool is_dy;
   bool is_wjets;
+  bool is_qcd;
   bool is_alps;
   bool is_azh;
   int i_mu_r = 0, i_mu_f = 0;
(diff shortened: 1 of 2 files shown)


jabuschh:
Here's how it looked before:

MCScale_QCD_old

And here's how it looks now with the correct indexing:

MCScale_QCD

mschrode:
Hi @jabuschh , thanks! Don't we also have pythia QCD samples? I guess those would have the other indexing?

jabuschh:
> Hi @jabuschh , thanks! Don't we also have pythia QCD samples? I guess those would have the other indexing?

Ah yes, you're right. If I see it correctly just the HT-binned ones were produced with MadGraph + Pythia as the pT-binned ones were produced with Pythia only.
I will change the flag accordingly.

PR #1650, opened 2022-09-19
PSWeights module: fix public inheritance / constructor arg to be able to force use of dummy weights

MatthiesC:
[only compile]
Fixed missing `public` that caused compilation error "`error: 'uhh2::AnalysisModule' is an inaccessible base of 'PSWeights'`"


MatthiesC:
Second commit: Add constructor argument to be able to force use of dummy weights if e.g. sample has no PS weights stored

PR #1651, opened 2022-09-20
update PSWeights module

MatthiesC:
[only compile]

Updates PSWeights module

- made all event handles private
- better naming scheme for all handles/weights; also corrected the naming of `g2qg` to `q2qg` and `g2xg` to `x2xg`; now only using sqrt2, 2, or 4 in the names for easier reading (no 025, 05 etc. anymore)
- includes now also the cNS weights for the splitted weights
- new option to skip writing the splitted weights to your AnalysisTree if you don't need those weights, see new constructor argument

### DIFF
--- a/common/include/PSWeights.h
+++ b/common/include/PSWeights.h
@@ -2,53 +2,79 @@
 #include "UHH2/core/include/AnalysisModule.h"
 #include "UHH2/core/include/Event.h"
 
-
+// See this TWiki for explanation of all the different weights:
+// https://twiki.cern.ch/twiki/bin/view/CMS/HowToPDF#Parton_shower_weights
+// You have the option to turn off writing the splitted weights to your AnalysisTree if you don't need those
 class PSWeights: public uhh2::AnalysisModule {
 public:
-  explicit PSWeights(uhh2::Context & ctx, const bool sample_without_ps_weights_ = false);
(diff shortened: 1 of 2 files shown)


mschrode:
Hi @MatthiesC , thanks! Why is the name `...isr...mu_r`? Isn't the alpha_s in the PS varied, not mu_r?

MatthiesC:
I took the naming from the TWiki: https://twiki.cern.ch/twiki/bin/view/CMS/HowToPDF#Parton_shower_weights
In particular from this image (I also checked it in the ROOT files on my own): https://twiki.cern.ch/twiki/pub/CMS/HowToPDF/pythia8_weights_ul.png
There is a difference between the splitted muR and cNS weights. The inclusive weights have also muR in the name, but I dropped it here since there is no other inclusive weight than the muR one.

mschrode:
Ah, OK, thanks for the clarification. I checked the [description of the variations](https://twiki.cern.ch/twiki/bin/view/CMS/TopSystematics#Parton_shower_ISR_and_FSR_govern), and indeed: "The uncertainty is evaluated by varying the renormalization scale for QCD emissions in initial-state and final-state radiation (ISR and FSR), effectively varying the value of αS." So it is a mu_r variation actually

PR #1653, opened 2022-09-23
add ZPrimeToTT and HeavyHiggsToTT flags to MC Scale Variations

kdeleo:
[106X] Adding "ZPrimeToTT" and "HeavyHiggsToTT" flags to MC Scale Variations after checking the weights indexing in the LHE files

[ci skip]

### DIFF
--- a/common/include/MCWeight.h
+++ b/common/include/MCWeight.h
@@ -82,6 +82,9 @@ class MCScaleVariation: public uhh2::AnalysisModule {
   bool is_qcd_HTbinned;
   bool is_alps;
   bool is_azh;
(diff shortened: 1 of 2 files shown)


MatthiesC:
I wonder if it would not make sense at this point to add the option to define dataset name matches e.g. as a config parameter in the SFrame config XML and take that as input for the MCScaleVariation class. But this has been discussed already before, right?

kdeleo:
Yes it has been discussed here #1643

kdeleo:
@MatthiesC could you merge this PR for now?

PR #1654, opened 2022-10-27
[106X_v2] [MCWeight] Error propagation for Lepton SF

bonanomi:


bonanomi:
I have not included any special handling of crack electrons, assuming that everything remain as it is and people mask them in their analysis. If we want, we can include a check for which:

    if(err>1) err=0;

So that crack electrons (for which SF=1, hence no effect) do not enter the error propagation.

Since also in AZH we decided to mask crack electrons, my preference would be to leave this module as it is to maintain generality.

danielhundhausen:
Looks good to me. @mschrode @MatthiesC can you merge this?

PR #1655, opened 2022-11-10
[106X] TstarTstar MC scale variation handling

finnlabe:
[only compile]

Same change as in https://github.com/UHH2/UHH2/pull/1653 for my tstartstar signal samples.


finnlabe:
Fascinating... lets test again.

PR #1656, opened 2022-11-23
Update functionality of read entry module [106X]

anmalara:
[ci skip]
[only  compile]


mschrode:
Hi @anmalara , what is the status here? Should we merge?

anmalara:
For me yes. not sure if @mschrode or @stalbrec were testing other options

stalbrec:
I have not gotten around to test this yet, but the stuff i tested (not extensively) earlier looked similar. I don't have any objections against this :)

PR #1658, opened 2022-11-25
[all] change calculation of cos(θ*)

jabuschh:
I slightly changed the calculation of the cos(θ*) observable (mainly for future reference).
Before, it was calculated as the azimuthal angle with respect to the beam line.
However, as I've found out during my reinterpretation studies of TOP-20-001, the "correct" (or more commonly used) definition is with respect to the ttbar system.

See the second commit and [these slides](https://indico.desy.de/event/36778/#2-henrik) for more details.

[only compile]


jabuschh:
I just saw that the tests failed, I will fix it and retry

jabuschh:
Done. Now it should compile. I simply forgot a ```.Vect()```.
Could you restart the compiling? Thanks!

PR #1662, opened 2023-02-20
add missing 2016 UL map elements in year_str_map_simple

danielhundhausen:


MatthiesC:
I guess this can be merged, right?
@danielhundhausen Might make sense to specify `[only @ compile]` (without the `@`) next time in the PR message to not trigger the entire testing which is only relevant if we do changes to ntuple generating code.

PR #1669, opened 2023-07-03
[Muon RECO SF] Add muon reco SF .root files

bonanomi:
This PR includes Muon RECO SF in `UHH2-data/muon_SFs_UL` as so far we only had ID and ISO SFs, but no RECO SFs. Given that the structure of these `.root` file reflects the one of the ID+ISO SFs, no modification to `MCMuonScaleFactor` should be needed.

SFs taken from: https://twiki.cern.ch/twiki/bin/view/CMS/MuonUL2018#RECO_efficiency


stalbrec:
Hi Matteo, i just had a quick look and noticed that the root files you added are actually html files. I guess you curled them directly and they are behind SSO?

bonanomi:
> Hi Matteo, i just had a quick look and noticed that the root files you added are actually html files. I guess you curled them directly and they are behind SSO?

🤦‍♂️ Thanks for noticing this Steffen, my bad. Let me replace them with the proper `.root` files

danielhundhausen:
@stalbrec can you merge this for us?

PR #1674, opened 2024-10-13
[RunII_106X_v2] Fixed error propgation formula for MC<Muon/Elec>ScaleFactor classes

MatthiesC:
[only compile]

This updates the changes made in https://github.com/UHH2/UHH2/pull/1654. The formula implemented in that PR was still not correct but the effect of this error was kinda low since most of the time `weight` is approx. one.


MatthiesC:
Pipeline failed with:
```
http://linuxsoft.cern.ch/epel/7/x86_64/repodata/repomd.xml: [Errno 14] HTTP Error 404 - Not Found
Cleaning up project directory and file based variables 00:01
ERROR: Job failed: command terminated with exit code 1
```
I guess this is due to EL7 being retired now? The code definitely compiles in my EL7 singularity container.

